PR #2, opened 2020-07-01
Implement event dispatching

stefanprodan:
Changes:
- Port notifier package from Flagger 
- Implement events dispatching for Slack, MS Teams, Discord, Rocket chat
- Implement events forwarding for `webhook` type


main.go
```diff
@@ -45,20 +46,23 @@ func init() {
 
 func main() {
 	var (
+		eventsAddr           string
 		metricsAddr          string
 		enableLeaderElection bool
 		concurrent           int
 		logJSON              bool
 	)
 	flag.StringVar(&metricsAddr, "metrics-addr", ":8484", "The address the metric endpoint binds to.")
```
hiddeco, line 55: source-controller has `:8080`, kustomize-controller `:8282`, this `:8484`?

stefanprodan: Yes, because I run them all locally.

stefanprodan: Please make sure helm-controller uses its own port.

hiddeco: Wouldn't it be better for them to all be the same so things are predictable across components?

stefanprodan: All of them have the same port name and Prometheus uses the annotations we setup in the deployment, that port is not intended for a user to call into. We could use the same port and set the current ones in the makefile under run command if you want.

hiddeco: > We could use the same port and set the current ones in the makefile under run command if you want.

I think having the same defaults for "standard flags" across components (even if they are not user facing) is the right thing to do.

stefanprodan: 8080 for HTTP API and 9090 for Prom?

hiddeco: kubebuilder automatically assigns `:8080` to Prom so I think the opposite.

stefanprodan: OK

controllers/alert_controller.go
```diff
@@ -18,13 +18,16 @@ package controllers
...
 
 	"github.com/go-logr/logr"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
-	notificationv1alpha1 "github.com/fluxcd/notification-controller/api/v1alpha1"
+	apiv1 "github.com/fluxcd/notification-controller/api/v1alpha1"
```
hiddeco, line 30: Given the fact that it is our own API, I think the `api` prefix is redundant (and not adding any additional context) and it can be just `v1`.

PR #22, opened 2020-07-20
Upgrade CRDs to apiextensions.k8s.io/v1

stefanprodan:
Ref: https://github.com/fluxcd/toolkit/issues/111


stefanprodan:
I prefer to run the e2e tests on the latest stable version instead of the lowest supported. If you want to test on all versions then we should do a matrix out of all Kubernetes supported versions e.g. 1.16/17/18

hiddeco:
> If you want to test on all versions then we should do a matrix out of all Kubernetes supported versions e.g. 1.16/17/18

Ack. In case of a matrix, I would only run all versions for `master` builds.

stefanprodan:
> Ack. In case of a matrix, I would only run all versions for master builds.

That's not possible with GH Actions, we could duplicate the whole workflow for PRs but then we need to keep them in sync.

hiddeco:
@stefanprodan let's record this in an issue and discuss details there.

Issue #26, opened 2020-07-29
Github Notifier


phillebaba:
The Github Notifier given an event containing revision metadata it updates the commit status in a Github repository.

The notifier would be configured in a similar manner as the other notifiers. With some small differences. The address given in the provider would have be the address of the Github repository. Aditionally the provider requires a secret ref to be given the personal access token to communicate with Github. The Alert would need to explicitly specify the resources to react upon as the source repository they are reconciled with have to match the repository of the provider. Another solution would be if we could get the source repository of the involved object and match it to a provider with the same repository, but that would go against the current behavior of the notifier.
```yaml
apiVersion: notification.fluxcd.io/v1alpha1
kind: Provider
metadata:
  name: podinfo-gh-status
  namespace: gitops-system
spec:
  type: github-commit-status
  address: https://github.com/stefanprodan/podinfo # repo URL
  secretRef:
    name: github-token # GH token with 'repo:status' scope
---
apiVersion: v1
kind: Secret
metadata:
  name: github-token
  namespace: gitops-system
data:
  token: <token>
---
apiVersion: notification.fluxcd.io/v1alpha1
kind: Alert
metadata:
  name: podinfo-status-update
  namespace: gitops-system
spec:
  providerRef: 
    name: podinfo-gh-status 
  eventSeverity: info
  eventSources:
    - kind: Kustomization
      name: webapp-backend
---
apiVersion: kustomize.fluxcd.io/v1alpha1
kind: Kustomization
metadata:
  creationTimestamp: null
  name: webapp-backend
  namespace: gitops-system
spec:
  dependsOn:
  - webapp-common
  healthChecks:
  - kind: Deployment
    name: backend
    namespace: webapp
  interval: 10m0s
  path: ./deploy/webapp/backend
  prune: true
  sourceRef:
    kind: GitRepository
    name: webapp
  timeout: 2m0s
  validation: client
status: {}
```

An detail in this implementation is that if the Kustomization resource and the source repoisitory sepcified are not the same. The event revision relates to the source repository given in the Kustomization resource, and not the repository the Kustomization resource originates from.

stefanprodan:
Fixed in #27

PR #27, opened 2020-07-29
Implement GitHub commit status notifier

phillebaba:
Implements #26

### DIFF
--- a/internal/notifier/github.go
+++ b/internal/notifier/github.go
@@ -68,7 +68,8 @@ func (g *GitHub) Post(event recorder.Event) error {
 		return err
 	}
 
-	desc := fmt.Sprintf("%v %v %v", event.InvolvedObject.Kind, event.InvolvedObject.Name, event.Reason)
+	reason := strings.Join(split(event.Reason), " ")
+	desc := fmt.Sprintf("%v %v %v", event.InvolvedObject.Kind, event.InvolvedObject.Name, reason)
 	status := &github.RepoStatus{
 		State:       &state,
 		Context:     github.String("cd/flux"),
--- a/internal/notifier/github.go
+++ b/internal/notifier/github.go
@@ -68,11 +68,11 @@ func (g *GitHub) Post(event recorder.Event) error {
 		return err
 	}
 
-	reason := strings.Join(split(event.Reason), " ")
-	desc := fmt.Sprintf("%v %v %v", event.InvolvedObject.Kind, event.InvolvedObject.Name, reason)
+	githubCtx := fmt.Sprintf("%v/%v/%v", event.InvolvedObject.Kind, event.InvolvedObject.Namespace, event.InvolvedObject.Name)
+	desc := strings.Join(split(event.Reason), " ")
 	status := &github.RepoStatus{
 		State:       &state,
-		Context:     github.String("cd/flux"),
+		Context:     &githubCtx,
 		Description: &desc,
 	}
 
--- a/internal/notifier/github.go
+++ b/internal/notifier/github.go
@@ -105,7 +105,7 @@ func parseGithubAddress(addr string) (string, string, error) {
 func parseRevision(rev string) (string, error) {
 	comp := strings.Split(rev, "/")
 	if len(comp) < 2 {
-		return "", errors.New("Revision string format incorrect")
+		return "", fmt.Errorf("Revision string format incorrect: %v", rev)
 	}
 
 	return comp[1], nil
--- a/internal/notifier/github.go
+++ b/internal/notifier/github.go
@@ -68,8 +68,10 @@ func (g *GitHub) Post(event recorder.Event) error {
 		return err
 	}
 
-	githubCtx := fmt.Sprintf("%v/%v/%v", event.InvolvedObject.Kind, event.InvolvedObject.Namespace, event.InvolvedObject.Name)
+	githubCtx := fmt.Sprintf("%v/%v", event.InvolvedObject.Kind, event.InvolvedObject.Name)
+	githubCtx = strings.ToLower(githubCtx)
 	desc := strings.Join(split(event.Reason), " ")
+	desc = strings.ToLower(desc)
 	status := &github.RepoStatus{
 		State:       &state,
 		Context:     &githubCtx,
--- a/internal/notifier/github.go
+++ b/internal/notifier/github.go
@@ -1,0 +1,120 @@
+/*
+Copyright 2020 The Flux CD contributors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package notifier
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net/url"
+	"strings"
+	"time"
+
+	"github.com/fluxcd/pkg/recorder"
+	"github.com/google/go-github/v32/github"
+	"golang.org/x/oauth2"
+)
+
+type GitHub struct {
+	Owner  string
+	Repo   string
+	Client *github.Client
+}
+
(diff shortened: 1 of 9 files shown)


internal/notifier/github.go
```diff
@@ -0,0 +1,120 @@
...
+	rev, err := parseRevision(revString)
+	if err != nil {
+		return err
+	}
+	state, err := toGitHubState(event.Severity)
+	if err != nil {
+		return err
+	}
+
+	desc := fmt.Sprintf("%v %v %v", event.InvolvedObject.Kind, event.InvolvedObject.Name, event.Reason)
```
stefanprodan, line 71: We need to split the reason into words and make them lower case, here is a function you can copy/paste https://github.com/fatih/camelcase/blob/master/camelcase.go#L47

phillebaba: Right now the reason seems to be the same as the severity. I have only seen the values "info" and "error" but I will implement the function in case it changes.

phillebaba:
This is the resulting commit status of a successful deployment. 

image

internal/notifier/github.go
stefanprodan: > Right now the reason seems to be the same as the severity

This is a bug, the reason should be one of: https://github.com/fluxcd/kustomize-controller/blob/master/api/v1alpha1/condition_types.go#L56-L90

stefanprodan: Fixed in https://github.com/fluxcd/kustomize-controller/pull/76

stefanprodan:
@phillebaba can we set the icon to the [FluxCD one](https://github.com/fluxcd/toolkit/blob/master/docs/_files/flux-icon%402x.png) the public URL is `https://raw.githubusercontent.com/fluxcd/toolkit/master/docs/_files/flux-icon%402x.png`.

phillebaba:
@stefanprodan It seems like it is not possible to change the icon when using a PAT to authenticate. Github will fetch the avatar of the user who owns the PAT and use that to authenticate. After reading a bit, it seems like it is only possible to use custom avatars through Github applications. In that case each end user would have to create their own application instead.

phillebaba:
We still need to figure out the context so that it is not overriden every time. If you look att the e2e test the context is "e2e / kind (pull_request)" without a message. This allows Github Actions to post the status from multiple actions on a single sha. 

In my mind, for each Kustomization and HelmRelease has to events, that reconcile and the health check. Either we merge these two together or they become a separate event. Either way each event needs its own unique context. My suggestion is something similar to "cd/flux/{{ .Kind }}/{{ .Name }}" maybe even remove the "cd/flux" prefix.

stefanprodan:
@phillebaba let's go with `{{ .Kind }}/{{ .Name }}` and remove cd/flux as the label shouldn't be too long.

stefanprodan:
@phillebaba please rebase with master and reinstall the toolkit, to clean your cluster do `tk uninstall --crds`. The reason should work now with the latest toolkit release.

phillebaba:
I will try to get it done later today. The CRD changes should not affect this functionality too much so it should be a quick job.

stefanprodan:
@phillebaba can you please post here a print screen with new format?

phillebaba:
I added the namespace of the resource as part of the context, and as you can see the text it overflowing.

image

Its sucks that there is no way of getting a detailed view of the status, but its fine as I use the information through API calls anyways.

stefanprodan:
I don't think we should add the namespace, all kustomization work if they are in the same namespace as the control plane.

phillebaba:
Ok so this would only be an issue for user using the same repo for multiple control planes.

Additionally this will be an issue for remote ref of Kustomization resources. I created my setup according to the tutorial, so I have a fleet-infra repo, that contains Kustomization resources for a podinfo app. These resource are deployed in the cluster.
```
apiVersion: kustomize.toolkit.fluxcd.io/v1alpha1
kind: Kustomization
metadata:
  creationTimestamp: null
  name: webapp-backend
  namespace: gitops-system
spec:
  dependsOn:
  - webapp-common
  healthChecks:
  - kind: Deployment
    name: backend
    namespace: webapp
  interval: 10m0s
  path: ./deploy/webapp/backend
  prune: true
  sourceRef:
    kind: GitRepository
    name: webapp
  timeout: 2m0s
  validation: client
status: {}
---
apiVersion: source.toolkit.fluxcd.io/v1alpha1
kind: GitRepository
metadata:
  creationTimestamp: null
  name: webapp
  namespace: gitops-system
spec:
  interval: 30s
  ref:
    branch: master
  url: https://github.com/stefanprodan/podinfo
status: {}
```

The event received will contain the commit id for the `podinfo` repo and not the `fleet-infra` repo. In my case this is `d3865f924736fe5d4c52bb9f8db86e46cfd1fe52` [which is this commit](https://github.com/stefanprodan/podinfo/commit/d3865f924736fe5d4c52bb9f8db86e46cfd1fe52). If we follow this notification behavior multiple instances would update the same commit sha. 

Maybe we should remove the namespace and add a instance field (dev, qa, prod)? `{{ .Instance }}/{{ .Kind }}/{{ .Name }}` to differentiate between toolkit instances.

stefanprodan:
To differentiate between toolkit instances users could use a prefix/suffix for Kustomization naming, e.g. `webapp-prod`, `webapp-staging`. Should we change the get started guide and use such names?

phillebaba:
I guess all of this depends on the multi tenancy discussion.
https://github.com/fluxcd/toolkit/discussions/93

If the guidelines would be to prefix/suffix with environment name even when you have separate clusters for each environment, then I am fine with it. Then I would just assume uniqueness in the name in my implementation, so that the controller does not override statuses.

I am still a bit confused with the metadata though. My understanding, which turns out is wrong, was that the event was given the commit id for the origin of the `HelmRelease` or `Kustomization`. It turns out it the commit id refers to where the Kustomization points towards.
Untitled Diagram(1)

I want to get the CommitID 1 to the notification controller and set the status in that repo, which is not happening now. This was my original idea for how it would work, do you agree with this?

stefanprodan:
The feel-infra repo contains GitRepositories and Kustomizations. When you create an alert it should not be for the fleet-infra but for the repository that's referenced in GitRepository. The only exception to this is the `gitops-system` repo+kustomization that synchronises all the repos and kustomizations.

phillebaba:
Ok i guees i didnt understand the future workflow. It is a bit different from current flux which is more single repo centered. I am currently on vaccation but will look into this later when i get back.

phillebaba:
So I made some small changes, that I think are now inline with the way of working. I guess I had to play around a bit with gitops toolkit to understand some of your pointers.

The github notifier now works with multiple kustomizations in multiple git repositories. All that is required to receive a git status is to make sure that the provider github repo matches the git source of the kustomization source.
```yaml
apiVersion: source.toolkit.fluxcd.io/v1alpha1
kind: GitRepository
metadata:
  creationTimestamp: null
  name: test-app
  namespace: gitops-system
spec:
  interval: 30s
  ref:
    branch: master
  url: https://github.com/phillebaba/test-app
status: {}
---
apiVersion: kustomize.toolkit.fluxcd.io/v1alpha1
kind: Kustomization
metadata:
  name: test-app
  namespace: gitops-system
spec:
  interval: 30s
  path: ./staging
  prune: true
  sourceRef:
    kind: GitRepository
    name: test-app
  timeout: 2m0s
  validation: client
status: {}
---
apiVersion: notification.toolkit.fluxcd.io/v1alpha1
kind: Provider
metadata:
  name: test-app
  namespace: gitops-system
spec:
  type: github
  channel: general
  address: https://github.com/phillebaba/test-app
  secretRef:
    name: github
---
apiVersion: notification.toolkit.fluxcd.io/v1alpha1
kind: Alert
metadata:
  name: test-app
  namespace: gitops-system
spec:
  providerRef:
    name: test-app
  eventSeverity: info
  eventSources:
    - kind: Kustomization
      namespace: gitops-system
      name: 'test-app'
```

This results with the following status in my test-app repo.
image

internal/notifier/github_test.go
```diff
@@ -0,0 +1,56 @@
...
+	/*"encoding/json"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"*/
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+/*func TestGithub_Post(t *testing.T) {
```
stefanprodan, line 29: Can you please remove the commented code?

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @phillebaba 🏅

PR #43, opened 2020-09-13
Add GitLab notifier

### LINKED ISSUE #40: GitLab commit status notifier
Similar to [GitHub](https://github.com/fluxcd/notification-controller/pull/27), we could implement commit status updates for GitLab.

### LINKED ISSUE #44: Alert EventSeverity has no default value
The documentation states that if EventSeverity is not set it will default to `info`. I cannot find any code that sets the default value.

```go
// Filter events based on severity, defaults to ('info').
// +kubebuilder:validation:Enum=info;error
// +optional
EventSeverity string `json:"eventSeverity,omitempty"`
```

Additionally the check for EventSeverity has an undocumented behavior. As I understand it an Alert will be sent either if the EventSeverity of the event matches the alert. Or if the Alert severity is of type `info`. If this is the intended behavior it should be documented that setting the severity to `info` will also include alerts with severity `error`.
https://github.com/fluxcd/notification-controller/blob/20f07c809644fb45aae76ac185ffa5b9d4d73386/internal/server/event_handlers.go#L80-L81

### LINKED ISSUE #45: Filter events sent by different types of notifiers
The kustomization controller limits the amount of events sent to the notification controller, to limit the amount of message sent by the chat notifiers (Slack, MS Teams, etc.). If this was not done the messages would be very verbose, as events would be sent every reconcile loop, even when nothing has changes.

On the other hand the git notifiers need to receive succeeded events if a deployment goes from unhealthy to healthy without the revision changing.

These two requirements are at odds with each other. To deal with this the chat notifiers need to filter out any succeeded events where the revision has not changed from the last sent event.

phillebaba:
Adds a notifier for GitLab that acts in a similar manner to the GitHub notifier.

Closes #40 
Closes #44 
Closes #45 

**The front page of the repository**
image

**Pipeline detailed view**
image

**Info on mouse hover**
image


phillebaba:
The latest commit should also close #44.

phillebaba:
Default value annotations was verified that it worked in kind 1.16.4

internal/notifier/discord.go
```diff
@@ -64,6 +64,11 @@ func NewDiscord(hookURL string, username string, channel string) (*Discord, erro
 
 // Post Discord message
 func (s *Discord) Post(event recorder.Event) error {
+	// Skip any update events
+	if status, ok := event.Metadata["commit_status"]; ok && status == "update" {
```
stefanprodan, line 68: Can we please make this into a function `IsCommitStatus` and reuse it in all providers.

stefanprodan (changes requested):

stefanprodan (approved):
LGTM

Thanks @phillebaba 🎖️

PR #53, opened 2020-09-30
Implement `fluxcd/pkg/meta/api` in APIs

stefanprodan:


controllers/alert_controller.go
```diff
@@ -52,16 +54,16 @@ func (r *AlertReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	init := true
 	for _, condition := range alert.Status.Conditions {
-		if condition.Type == v1alpha1.ReadyCondition && condition.Status == corev1.ConditionTrue {
+		if condition.Type == meta.ReadyCondition && condition.Status == corev1.ConditionTrue {
 			init = false
 			break
 		}
 	}
```
hiddeco, line 61: This can now theoretically be replaced by something like

```suggestion
	if c := meta.GetCondition(alert.Status.Conditions, meta.ReadyCondition); c != nil {
		if c.Status == corev1.ConditionTrue {
			init = false
		}
	}
```

controllers/provider_controller.go
```diff
@@ -52,16 +54,16 @@ func (r *ProviderReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	init := true
 	for _, condition := range provider.Status.Conditions {
-		if condition.Type == v1alpha1.ReadyCondition && condition.Status == corev1.ConditionTrue {
+		if condition.Type == meta.ReadyCondition && condition.Status == corev1.ConditionTrue {
 			init = false
 			break
 		}
 	}
```
hiddeco, line 61: This can now theoretically be replaced by something like

```suggestion
	if c := meta.GetCondition(provider.Status.Conditions, meta.ReadyCondition); c != nil {
		if c.Status == corev1.ConditionTrue {
			init = false
		}
	}
```

controllers/receiver_controller.go
```diff
@@ -67,7 +69,7 @@ func (r *ReceiverReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 	isReady := false
 	for _, condition := range receiver.Status.Conditions {
```
hiddeco, line 71:
```suggestion
	if c := meta.GetCondition(receiver.Status.Conditions, meta.ReadyCondition); c != nil {
		isReady = c.Status == corev1.ConditionTrue
	}
```

hiddeco (approved):
:medal_sports:

PR #54, opened 2020-09-30
Promote the notification.toolkit.fluxcd.io API to v1beta1

stefanprodan:
This PR promotes the `notification.toolkit.fluxcd.io` API to v1beta1 and removes support for v1alpha1.

Going forward, changes to the API will be accompanied by a conversion mechanism. With this release the API becomes more stable but while in beta phase there are no guarantees about backwards compatibility between beta releases.

Ref: https://github.com/fluxcd/toolkit/issues/238


hiddeco (commented):
Missed a couple of mentions in `api/v1beta1/doc.go` and `controllers/suite_test.go`

Screenshot from 2020-09-30 17-14-45

hiddeco (approved):
:satellite:

PR #56, opened 2020-10-05
Update default branch and namespace

stefanprodan:
Changes:
- Set `main` as default branch
- Remove  `gitops-system` namespace refs

Ref: https://github.com/fluxcd/toolkit/issues/152

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
   push:
     branches:
-      - master
+      - main
 
 jobs:
   kind:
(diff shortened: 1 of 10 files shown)


hiddeco (commented):
How much value has the reference to the namespace in this document? For the helm-controller, I simply removed the references as from a controller context it offers little value over the cost of maintenance.

When this is merged, and we have switched to `main`, we need to edit the links in the [existing releases](https://github.com/fluxcd/notification-controller/releases) as they refer to the `master` branch.

stefanprodan:
How much value has the reference to the namespace in the documents?

To showcase the namespace selector usage

hiddeco:
> To showcase the namespace selector usage

That can also be done with a dummy namespace (at present it refers to the same namespace as the resource itself is in).

hiddeco (approved):

Issue #61, opened 2020-10-16
Add http(s) proxy support


dbluxo:
To send notifications behind a proxy, we would need the ability to configure a http(s) proxy.

```
{"level":"error","ts":"2020-10-16T14:59:37.134Z","logger":"event-server","msg":"failed to send notification","object":"gotk-system/example-app","kind":"Kustomization","error":"postMessage failed: POST https://webhook-url giving up after 5 attempt(s): Post \"https://webhook-url\": context deadline exceeded (Client.Timeout exceeded while awaiting headers)"}
```

PR #62, opened 2020-10-17
Add support for http(s) proxy when sending alerts

stefanprodan:
Changes:
- Add `Proxy` optional field to Provider API
- Add URL prefix validation to CRDs
- Implement http(s) proxy for alerts

Proxy example:
```yaml
kind: Provider
metadata:
  name: slack
  namespace: default
spec:
  type: slack
  channel: general
  address: https://hooks.slack.com/services/YOUR/SLACK/WEBHOOK
  proxy: https://proxy.corp:8080
```

Fix: #61

### DIFF
--- a/api/v1beta1/provider_types.go
+++ b/api/v1beta1/provider_types.go
@@ -42,6 +42,10 @@ type ProviderSpec struct {
 	// +optional
 	Address string `json:"address,omitempty"`
 
+	// HTTP(S) address of the proxy
+	// +optional
+	Proxy string `json:"proxy,omitempty"`
+
 	// Secret reference containing the provider webhook URL
 	// +optional
 	SecretRef *corev1.LocalObjectReference `json:"secretRef,omitempty"`
--- a/api/v1beta1/provider_types.go
+++ b/api/v1beta1/provider_types.go
@@ -38,11 +38,15 @@ type ProviderSpec struct {
 	// +optional
 	Username string `json:"username,omitempty"`
 
-	// HTTP(S) webhook address of this provider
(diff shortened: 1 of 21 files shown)


hiddeco (commented):
Given the existence of some other (low level) proxy protocols than HTTP/S (e.g. SOCKS5), I would prefix the field names with the protocol to guarantee we are not shooting ourselves in the foot right now, and leave room for other implementations later (if any).

config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml
```diff
@@ -52,6 +52,9 @@ spec:
               channel:
                 description: Alert channel for this provider
                 type: string
+              proxy:
+                description: HTTP(S) address of the proxy
```
hiddeco, line 56:
```suggestion
                description: HTTP/S address of the proxy
```

api/v1beta1/provider_types.go
```diff
@@ -42,6 +42,10 @@ type ProviderSpec struct {
 	// +optional
 	Address string `json:"address,omitempty"`
 
+	// HTTP(S) address of the proxy
```
hiddeco, line 45:
```suggestion
	// HTTP/S address of the proxy
```

(+ a validation rule to only accept `http://` and `https://`)

hiddeco (commented):
On second thoughts, we can require a protocol to be defined and make decisions based on that, if it ever gets to supporting other things.

hiddeco (approved):

PR #67, opened 2020-10-27
Use reconcile request annotation

stefanprodan:
Changes:
- update `fluxcd/pkg/apis/meta` to v0.1.0
- update `fluxcd/pkg/runtime` to v0.1.2
- trigger reconciliation with `reconcile.fluxcd.io/requestedAt` annotation

Ref: https://github.com/fluxcd/pkg/issues/39


hiddeco (approved):
LGTM :100:

PR #73, opened 2020-11-07
Add bitbucket notifier

### LINKED ISSUE #65: Add status support for Bitbucket
We should support Bitbucket in the same way that we support GitHub and GitLab.

[ktrysmt/go-bitbucket](https://github.com/ktrysmt/go-bitbucket) is probably the best option for a client to use with Bitbucket.

phillebaba:
Add Bitbucket notifier in the same way as GitHub and GitLab.

Fixes #65


stefanprodan (changes requested):
Please update the spec docs

stefanprodan (approved):
LGTM

Thanks @phillebaba 🍸

Issue #72, opened 2020-11-06
leaving channel unset silently breaks Slack provider


edef1c:
The `channel` field is marked optional, and leaving it out passes validation.
I would expect this to implicitly use the default channel of the webhook.
Checking the status of the `Provider` and `Alert` resources reveals nothing, and the notification controller appears healthy.
```
edef@jaguar ~/s/m/p/e/p/gke ((f478031a…))> kubectl describe -n flux-system provider/slack alert/slack svc/notification-controller
Name:         slack
Namespace:    flux-system
Labels:       kustomize.toolkit.fluxcd.io/checksum=de849d6f08587510d1a21f9a7adf9de6ed889e28
              kustomize.toolkit.fluxcd.io/name=platform-prod-gke
              kustomize.toolkit.fluxcd.io/namespace=flux-system
Annotations:  API Version:  notification.toolkit.fluxcd.io/v1beta1
Kind:         Provider
Metadata:
  Creation Timestamp:  2020-11-06T15:38:36Z
  Generation:          1
  Resource Version:    5091913
  Self Link:           /apis/notification.toolkit.fluxcd.io/v1beta1/namespaces/flux-system/providers/slack
  UID:                 0db75803-2f28-4297-96e9-0d6178eab6fb
Spec:
  Secret Ref:
    Name:  slack-webhook
  Type:    slack
Status:
  Conditions:
    Last Transition Time:  2020-11-06T15:38:36Z
    Message:               Initialized
    Reason:                Initialized
    Status:                True
    Type:                  Ready
Events:                    <none>


Name:         slack
Namespace:    flux-system
Labels:       kustomize.toolkit.fluxcd.io/checksum=de849d6f08587510d1a21f9a7adf9de6ed889e28
              kustomize.toolkit.fluxcd.io/name=platform-prod-gke
              kustomize.toolkit.fluxcd.io/namespace=flux-system
Annotations:  API Version:  notification.toolkit.fluxcd.io/v1beta1
Kind:         Alert
Metadata:
  Creation Timestamp:  2020-11-06T15:38:36Z
  Generation:          1
  Resource Version:    5091911
  Self Link:           /apis/notification.toolkit.fluxcd.io/v1beta1/namespaces/flux-system/alerts/slack
  UID:                 fd502ac1-cdd6-44ff-ac7a-c4684f547086
Spec:
  Event Severity:  info
  Event Sources:
    Kind:  GitRepository
    Name:  *
    Kind:  Kustomization
[...]

PR #74, opened 2020-11-09
Add validation for providers and alerts

stefanprodan:
Validate providers and alerts and report the validation errors in the status sub-resource.

Fix: #72


phillebaba (commented):
Really nice job with the validation!

Should we skip alerts and providers in the event handler in the same way we do with suspended alerts?
https://github.com/fluxcd/notification-controller/blob/6c855a22bc0bdee778eff9f9ad8e2c17aa48f525/internal/server/event_handlers.go#L68

phillebaba (approved):
LGTM

PR #80, opened 2020-11-16
Change code to use subgroup in gitlab

Joda89:
When I use the subgroup of gitlab I get an error "Incorrectly formatted git address"
the id on gitlab can use more than two block
I changed the code to be able to handle two or more blocks in the ID
Can you tell me if this is right for you?

### DIFF
--- a/internal/notifier/util.go
+++ b/internal/notifier/util.go
@@ -32,14 +32,9 @@ func parseGitAddress(s string) (string, string, error) {
 		return "", "", nil
 	}
 
-	path := strings.TrimLeft(u.Path, "/")
-	comp := strings.Split(path, "/")
-	if len(comp) != 2 {
-		return "", "", fmt.Errorf("Incorrectly formatted git address: %v", s)
-	}
-
+	id := strings.TrimLeft(u.Path, "/")
+	id = strings.TrimSuffix(id, ".git")
 	host := fmt.Sprintf("https://%s", u.Host)
-	id := comp[0] + "/" + strings.TrimSuffix(comp[1], ".git")
 	return host, id, nil
 }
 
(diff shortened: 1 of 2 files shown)


stefanprodan:
@Joda89 please run `make test` and commit changes, then squash your commits into a single one. Thanks!

Joda89:
@stefanprodan It's done

phillebaba:
@Joda89 could you also add a test in `util_test.go` for a GitLab url with a subgroup. You should be able to just copy one of the existing tests.

Joda89:
@phillebaba it's done

phillebaba:
Sorry for not thinking about subgroups in GitLab, there are always small detail differences between the different GitHub providers. Thank you for taking the time and fixing it. :rose: 

Now when I think about it I don't really remember why I thought it was a good idea to split the path into components.
Could you change the code to just clean up the id value and return it?
```golang
func parseGitAddress(s string) (string, string, error) {
	u, err := giturls.Parse(s)
	if err != nil {
		return "", "", nil
	}

	id := strings.TrimLeft(u.Path, "/")
	id = strings.TrimSuffix(id, ".git")
	host := fmt.Sprintf("https://%s", u.Host)
	return host, id, nil
}
```

Each provider should be responsible for verifying the repo id they get so that it matches their client. GitHub needs to split it while GitLab does not. I can add that validation tonight to the different providers if you make my suggested changes so we merge this PR.

Joda89:
@phillebaba It's push 
tell me if it's good for you ?

phillebaba (approved):
LGTM :1st_place_medal: 

I will add some tests and then hopefully we can release notification-controller soon so the GitLab notifier.

Joda89:
thank

PR #82, opened 2020-11-17
Verify repository id in git notifiers

phillebaba:
The different git notifiers need repository ids in different formats.
On top of that some git providers support repositories in subgroups
while others do not support nesting. Each notifier needs test to verify
that they support their specific functionality.

### DIFF
--- a/internal/notifier/bitbucket.go
+++ b/internal/notifier/bitbucket.go
@@ -33,7 +33,7 @@ type Bitbucket struct {
 
 func NewBitbucket(addr string, token string) (*Bitbucket, error) {
 	if len(token) == 0 {
-		return nil, errors.New("Bitbucket token cannot be empty")
+		return nil, errors.New("bitbucket token cannot be empty")
 	}
 
 	_, id, err := parseGitAddress(addr)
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

Nice one, malformed ids will be reported on the provider status now!

internal/notifier/bitbucket.go
```diff
@@ -49,7 +50,7 @@ func NewBitbucket(addr string, token string) (*Bitbucket, error) {
 
 	comp = strings.Split(id, "/")
 	if len(comp) != 2 {
-		return nil, errors.New("Invalid bitbucket repository id")
+		return nil, fmt.Errorf("Invalid repository id '%s'", id)
```
bigkevmcd, line 53: minor: https://github.com/golang/go/wiki/Errors#errors (few cases of this in the code).

internal/notifier/github.go
```diff
@@ -35,20 +36,24 @@ type GitHub struct {
...
 	}
 
 	_, id, err := parseGitAddress(addr)
 	if err != nil {
 		return nil, err
 	}
 
+	comp := strings.Split(id, "/")
+	if len(comp) != 2 {
+		return nil, fmt.Errorf("Invalid repository id '%s'", id)
```
bigkevmcd, line 49: Usually nicer to use %q for these sorts of strings.

phillebaba:
@bigkevmcd Thanks for the tips out error message formatting!

@stefanprodan Should be ready to merge. Yep hopefully finding errors in the git url will be easier now.

PR #84, opened 2020-11-18
Send notification header identify requests as coming from the notification component.

bigkevmcd:
Add support for sending a Notification-Controller header from the forwarder notifier.

This means that it can be identified in upstream requests.

### DIFF
--- a/internal/notifier/forwarder.go
+++ b/internal/notifier/forwarder.go
@@ -21,8 +21,15 @@ import (
 	"net/url"
 
 	"github.com/fluxcd/pkg/recorder"
+	"github.com/hashicorp/go-retryablehttp"
 )
 
+// NotificationHeader is a header sent to identify requests from the
+// notification controller.
+const NotificationHeader = "gotk-component"
+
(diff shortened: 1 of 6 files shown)


internal/notifier/forwarder.go
```diff
@@ -40,7 +43,10 @@ func NewForwarder(hookURL string, proxyURL string) (*Forwarder, error) {
 }
 
 func (f *Forwarder) Post(event recorder.Event) error {
-	err := postMessage(f.URL, f.ProxyURL, event)
+	err := postMessage(f.URL, f.ProxyURL, event, func(req *retryablehttp.Request) {
+		req.Header.Set("Notification-Controller", event.ReportingController)
```
stefanprodan, line 47: Please extract the header name into a variable and rename it to `gotk-component`

phillebaba (commented):
I think this looks good, but I will let @stefanprodan have last say if he has any other comments.

stefanprodan (approved):

PR #86, opened 2020-11-22
Add Azure DevOps provider

phillebaba:
Add Azure DevOps provider as the changes to the source-controller should now support Azure DevOps. After this we should have covered the major git providers out there that have a commit status api.


phillebaba:
After getting the source-controller changes merged I should have some time over to finally update the docs with examples of the different git providers. I will get that done after this is merged.

Here is an example of how it looks in Azure DevOps.
image
image

stefanprodan (approved):
LGTM

Thanks @phillebaba 🍰

PR #88, opened 2020-11-24
Fix "Notification API reference" title

gertux:
Signed-off-by: Gert Dewit <gert@hobbiton.be>


stefanprodan (approved):
LGTM

Thanks @gertux 🍰

Issue #95, opened 2020-12-04
Receivers not covered by --watch-all-namespaces


Nonymus:
While setting up a multi-tenant testbed, I fail to create a webhook to sync the tenant's GitRepository in the tenant's namespace. Receiver for the flux-system namespace works fine.

Looking at https://github.com/fluxcd/notification-controller/blob/a7e3be032e0fa3dbf52425d65c00232b85bc4fe7/internal/server/receiver_handlers.go#L46 , it seems like notification-controller will only react on Receivers in its own namespace. Is this unintentional, or should we run a notficaiton-controller instance per namespace for Receivers?

PR #96, opened 2020-12-06
Lookup ready receivers in all namespaces

stefanprodan:
Lookup receivers in all namespaces filtered by readiness status.

Fix: #95


phillebaba (approved):
LGTM

PR #104, opened 2020-12-20
Upgrade controller-runtime to v0.7.0

phillebaba:
Ref: https://github.com/fluxcd/flux2/issues/613

This PR upgrades the `controller-runtime` dependency to `v0.7.0`,
including all changes required to make all wiring work again.

- Upgrade `runtime` to v0.6.0 to include `controller-runtime` changes.
- Loggers have been removed from the reconcilers and are now retrieved
  from the `context.Context` passed to the `Reconcile` method and
  downwards functions.
- Logger configuration flags are now bound to the flag set using
  `BindFlags` from `runtime/logger`, ensuring the same contract across
  GitOps Toolkit controllers, and the `--log-json` flag has been
  deprecated in favour of the `--log-encoding=json` default.
- The `ChangePredicate` from `runtime` has changed to a
  `ReconcilateAtChangedPredicate`, and is now chained with the
  `GenerationChangedPredicate` from `controller-runtime` using
  `predicate.Or`.
- The `leader-election-role` was changed, as leader election now works
  via the `coordination/v1` API.

Other notable changes:

- Use `DeletionTimestamp` of resource object to determine readiness.
- `source-controller` API dependency has been upgraded to `v0.6.0`
  which also uses `controller-runtime` `v0.7.0`.


controllers/alert_controller.go
```diff
@@ -86,15 +83,15 @@ func (r *AlertReconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	r.recordReadiness(ctx, alert)
 
 	return ctrl.Result{}, nil
 }
 
 func (r *AlertReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewControllerManagedBy(mgr).
 		For(&v1beta1.Alert{}).
-		WithEventFilter(predicates.ChangePredicate{}).
+		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicates.ReconcilateAtChangedPredicate{})).
```
phillebaba, line 94: Need to make sure this is the correct predicate.

go.mod
```diff
@@ -8,21 +8,22 @@ require (
 	github.com/fluxcd/notification-controller/api v0.5.0
 	github.com/fluxcd/pkg/apis/meta v0.5.0
 	github.com/fluxcd/pkg/recorder v0.0.6
-	github.com/fluxcd/pkg/runtime v0.4.0
+	github.com/fluxcd/pkg/runtime v0.5.1-0.20201217125143-9e1fe564d778
```
stefanprodan, line 11: We should use a stable release here

stefanprodan (approved):
LGTM

PR #105, opened 2021-01-06
Add hint about used "address" data key to the Provider CRD's SecretRef

joshmue:
When using a K8s `Secret` to store the webhook URL, the "address" data key has to be used. I know that this is mentioned in the docs and samples already, but when reading the spec first, this hint should improve clarity.

Please note if I missed anything.


stefanprodan (approved):
LGTM

Thanks @joshmue

PR #122, opened 2021-01-18
add service account to kustomize resources

willejs:
This PR adds a notification-controller SA and binds the required existing roles to it, and tells the deployment to use it.
This will enable us to move away from the default SA which has very permissive roles bound to it by other components.
This is mentioned in this issue https://github.com/fluxcd/flux2/issues/243

I have a couple of queries with regards to this. In the `clusterRoleBinding` that binds the manager-role to the service account, I need to specify the namespace the service account is in. If I do this, I am presuming it will mean I have to make changes in manifestgen in flux2? Additionally, I should add a patch into the default Kustomization rather than hardcoding the namespace in the roleBinding resource?

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -17,6 +17,7 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8080"
     spec:
+      serviceAccountName: notification-controller
       terminationGracePeriodSeconds: 10
       containers:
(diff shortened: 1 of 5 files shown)


stefanprodan:
Found a way to avoid namespace hardcoding without modifying kubebuider generated roles. Please see: https://github.com/fluxcd/flux2/pull/777

PR #124, opened 2021-01-19
Add object kind to messages

### LINKED ISSUE #120: Adding object kind to notifications
In some notification providers like Teams it is not always clear from which object type the notifications origin.

In Slack the originating controller is used as a username of the message, but in Teams the username is set by the webhook configuration.

Would it be possible to add the Kind of the causing object to the objName that is part of the message. 
E.g. as pre-/ or postfix like
`<Kind>/<name>.<namespace>`
or
`<name>.<namespace>.<kind>` (to stay in hostname notation)

(from https://github.com/fluxcd/flux2/issues/710 )

SomtochiAma:
This pull request adds the object kind of the source of the notification to the message.
The format is `<kind>/<name>.<namespaces>`
Fixes #120 

### DIFF
--- a/internal/notifier/discord.go
+++ b/internal/notifier/discord.go
@@ -91,7 +91,7 @@ func (s *Discord) Post(event recorder.Event) error {
 
 	a := SlackAttachment{
 		Color:      color,
-		AuthorName: fmt.Sprintf("%s/%s.%s", event.InvolvedObject.Kind, event.InvolvedObject.Name, event.InvolvedObject.Namespace),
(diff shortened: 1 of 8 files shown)


phillebaba:
Can we mimic the current implementation of the git notifications? In those providers the kind is already included and we lowercase all of the text.

```
 kustomization/apps-dev — reconciliation succeeded
```

SomtochiAma:
> Should we mimic the current implementation of the git notifications? In those providers the kind is already included and we lowercase all of the text.
> 
> ```
>  kustomization/apps-dev — reconciliation succeeded
> ```

@stefanprodan What do you think? We should definitely maintain a consistent format

stefanprodan:
Yes the kind should be lowercase, this is also consistent with kustomize-controller apply logs.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma

PR #127, opened 2021-01-20
Add generic webhook receiver for HMAC signing

### LINKED ISSUE #99: Allow signing of generic webhook provider requests
The HTTP endpoint should be able to ensure the provenance of incoming requests from the `generic` provider.

GitHub and GitLab use a shared-secret mechanism, where the outgoing requests have a header with an HMAC of the request body, the receiver can then sign the body it receives, and confirm the origin of the request.

_This should be optional_.

One proposal, is to add an additional signature field to the spec:

```yaml
spec:
 signature:
   secretRef:
       name: hmac
```

Another option would be to use the presence of an `hmac` key in the Secret referenced by `spec.secretRef`.

Some other considerations, sha1 or sha256? What should the header name be?

SomtochiAma:
Fixes #99

### DIFF
--- a/docs/spec/v1beta1/receiver.md
+++ b/docs/spec/v1beta1/receiver.md
@@ -38,15 +38,16 @@ Receiver types:
 
 ```go
 const (
-	GenericReceiver   string = "generic"
-	GitHubReceiver    string = "github"
-	GitLabReceiver    string = "gitlab"
-	BitbucketReceiver string = "bitbucket"
-	HarborReceiver    string = "harbor"
-	DockerHubReceiver string = "dockerhub"
-	QuayReceiver      string = "quay"
-	GCRReceiver       string = "gcr"
-	NexusReceiver     string = "nexus"
+	GenericReceiver     string = "generic"
+	GenericHMACReceiver string = "generic-hmac"
+	GitHubReceiver      string = "github"
+	GitLabReceiver      string = "gitlab"
+	BitbucketReceiver   string = "bitbucket"
+	HarborReceiver      string = "harbor"
+	DockerHubReceiver   string = "dockerhub"
+	QuayReceiver        string = "quay"
+  GCRReceiver         string = "gcr"
+  NexusReceiver       string = "nexus"
 )
 ```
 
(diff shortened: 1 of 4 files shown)


docs/spec/v1beta1/receiver.md
```diff
@@ -101,6 +102,67 @@ spec:
...
+kind: Receiver
+metadata:
+  name: generic-hmac-receiver
+  namespace: default
+spec:
+  type: generic-hmac
+  secretRef:
+    name: webhook-token
+  resources:
+    - kind: ImageRepository
```
stefanprodan, line 118:
```suggestion
    - kind: GitRepository
```

docs/spec/v1beta1/receiver.md
```diff
@@ -38,15 +38,16 @@ Receiver types:
...
-	NexusReceiver     string = "nexus"
+	GenericReceiver     string = "generic"
+	GenericHMACReceiver string = "generic-hmac"
+	GitHubReceiver      string = "github"
+	GitLabReceiver      string = "gitlab"
+	BitbucketReceiver   string = "bitbucket"
+	HarborReceiver      string = "harbor"
+	DockerHubReceiver   string = "dockerhub"
+	QuayReceiver        string = "quay"
+  GCRReceiver         string = "gcr"
```
stefanprodan, line 49: This needs some spacing

internal/server/receiver_handlers.go
```diff
@@ -110,6 +110,21 @@ func (s *ReceiverServer) validate(ctx context.Context, receiver v1beta1.Receiver
...
+			return fmt.Errorf("unable to read request body: %s", err)
+		}
+
+		err = github.ValidateSignature(r.Header.Get("X-Signature"), b, []byte(token))
+		if err != nil {
+			return fmt.Errorf("unable to validate signature: %s", err)
+		}
+
+		s.logger.Info(
+			"handling event from generic wehbook",
```
stefanprodan, line 125:
```suggestion
			"handling event from generic-hmac wehbook",
```

stefanprodan (approved):
LGTM

PR #128, opened 2021-01-21
Allow ImageUpdateAutomations in object refs

squaremo:
This adds ImageUpdateAutomation and ImagePolicy as kinds allowed by cross-namespace
object references here. That has two effects:

 - ImageUpdateAutomation and ImagePolicy objects can be the source of events; and,
 - ImageUpdateAutomation and ImagePolicy objects can be the target of webhook
   triggers.

Of these, the first is certainly desirable (e.g,. now you can post a Slack message when automation fails).

The second may be useful for ImageUpdateAutomation, though automations will more usually be triggered by ImageRepository objects changing; but in any case, doesn't hurt. It is not useful for ImagePolicy objects, but on balance, I think it's better to keep the type generic than to make a special case for ImagePolicy.

### DIFF
--- a/api/v1beta1/reference_types.go
+++ b/api/v1beta1/reference_types.go
@@ -24,7 +24,7 @@ type CrossNamespaceObjectReference struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 
 	// Kind of the referent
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Issue #137, opened 2021-02-08
Exclude messages from alerting


stefanprodan:
In some cases it is desirable to not send alerts if the message body matches some string e.g. https://github.com/fluxcd/flux2/issues/760

PR #138, opened 2021-02-08
Implement regex exclusions for alerts

stefanprodan:
This PR adds a option field called `exclusionList ` to the Alert API. 

Fix: #137

## Example usage

Skip alerting if the message matches a golang regex from the exclusion list:

```yaml
kind: Alert
metadata:
  name: flux-system
  namespace: flux-system
spec:
  providerRef: 
    name: on-call-slack
  eventSeverity: error
  eventSources:
    - kind: GitRepository
      name: flux-system
  exclusionList:
    - "waiting.*socket"
```

The above definition will not send alerts for transient Git clone errors like:

```
unable to clone 'ssh://git@ssh.dev.azure.com/v3/...', error: SSH could not read data: Error waiting on socket
```


api/v1beta1/alert_types.go
```diff
@@ -42,6 +42,10 @@ type AlertSpec struct {
 	// +required
 	EventSources []CrossNamespaceObjectReference `json:"eventSources"`
 
+	// A list of golang regex expressions to be used for excluding meesages
```
hiddeco, line 45:
```suggestion
	// A list of Golang regex expressions to be used for excluding messages.
	// https://golang.org/pkg/regexp/syntax/
```

internal/server/event_handlers.go
```diff
@@ -72,6 +73,17 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 				continue
 			}
 
+			// skip alert if the message matches a regex from the exclusion list
+			if len(alert.Spec.ExclusionList) > 0 {
+				for _, exp := range alert.Spec.ExclusionList {
+					if r, err := regexp.Compile(exp); err == nil {
```
hiddeco, line 79: Not sure if swallowing the error is the right thing to do, as regexp can be tricky to configure.

stefanprodan: Returning the error doesn't make sense, as the sender will retry loop and cause DDOS.

hiddeco: Log it?

hiddeco (approved):

PR #143, opened 2021-02-22
Use numeric UID for user in Dockerfile

funkypenguin:
Hey guys!

This PR simply specifies the UID for the "controller" user created by `Dockerfile`, to satisfy Kubernetes PSPs which enforce non-root containers. Without this change, the pod won't start, with this error:

```
  Warning  Failed     12s (x4 over 26s)  kubelet            Error: container has runAsNonRoot and image has non-numeric user (controller), cannot verify user is non-root
```

(Note, a matching PR has been submitted for source-controller, at https://github.com/fluxcd/source-controller/pull/297

Cheers!
D


stefanprodan:
Please see https://github.com/fluxcd/flux2/issues/829 We have no intention on hardcoding the user id inside the container.

PR #145, opened 2021-02-23
Unit tests for event forwarding

squaremo:
This adds some tests to check that events are forwarded as specified by alerts.

The controllers themselves aren't needed; but since the internal/server package refer directly to Kubernetes resources, the tests share prerequisites with controller tests. For that reason I have started by putting these tests in controllers/.

I have tried not to touch the server code itself, but it was necessary to change it to use a fresh http.ServerMux rather than the (global) default, so that the tests can create an event server each.

I've taken the opportunity to include a very minor tidy of bits of internal/server/.


stefanprodan (approved):
LGTM

Thanks @squaremo

PR #149, opened 2021-02-27
Provide the ability to send events to Google Chat

djexp:
By essentially replicating the implementations for Teams and Slack a specific implementation for Google Chat has been developed and tested.

The are two points of note:

1. I've had to filter out 'Progressing' events as these seemed to replace the previous message within Google Chat. I couldn't easily find out why. However the output in chat is very similar to Discord so why this system does not display these messages is a mystery?

2. I've included the event timestamp as 'meta-data' within the message. Other implementation have not done this, so I'm happy to remove it for consistency sake if that is preferred.

Finally. I am not a Go developer, this is actually my first time coding in it. I like it, it's very clean. But I cut my teeth in C, C++, Java and beyond. So my apologies for any novice mistakes.

### DIFF
--- a/internal/notifier/google_chat.go
+++ b/internal/notifier/google_chat.go
@@ -122,26 +122,21 @@ func (s *GoogleChat) Post(event recorder.Event) error {
 	})
 
 	// Meta-Data
-	kvfields := make([]GoogleChatCardWidget, 0, len(event.Metadata)+1)
-	kvfields = append(kvfields, GoogleChatCardWidget{
-		KeyValue: &GoogleChatCardWidgetKeyValue{
-			TopLabel:         "TIMESTAMP",
-			Content:          event.Timestamp.String(),
-			ContentMultiLine: false,
-		},
-	})
-	for k, v := range event.Metadata {
-		kvfields = append(kvfields, GoogleChatCardWidget{
-			KeyValue: &GoogleChatCardWidgetKeyValue{
-				TopLabel:         k,
-				Content:          v,
-				ContentMultiLine: false,
-			},
+	if len(event.Metadata) > 0 {
+		kvfields := make([]GoogleChatCardWidget, 0, len(event.Metadata))
+		for k, v := range event.Metadata {
+			kvfields = append(kvfields, GoogleChatCardWidget{
+				KeyValue: &GoogleChatCardWidgetKeyValue{
+					TopLabel:         k,
+					Content:          v,
+					ContentMultiLine: false,
+				},
+			})
+		}
+		sections = append(sections, GoogleChatCardSection{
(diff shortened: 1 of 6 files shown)


internal/notifier/google_chat.go
```diff
@@ -0,0 +1,161 @@
...
+		Title:    objName,
+		SubTitle: event.ReportingController,
+	}
+
+	sections := make([]GoogleChatCardSection, 0)
+
+	// Message
+	messageText := event.Message
+	if event.Severity == recorder.EventSeverityError {
+		messageText = fmt.Sprintf("<font color=\"#ff0000\">%s</font>", event.Message)
```
phillebaba, line 112: Is this a supported method of changing the text color or something that just works?

phillebaba (commented):
I see that you are setting a proxy URL, is this to deal with authentication? I am not to sure that this is the way to go as an official solution. Why can you not pass a token to the provider?

djexp:
In answer to both questions:

Firstly the proxy was added for consistency with all of the existing notification providers and is passed to the underlying HTTPClient during connection initialisation. I presume for accessing the internet for certain corporate networks. So it is not added for authentication, tokens are part of the URL as I believe they are for at least Slack and Teams.

Secondly, yes the embedded HTML markup is part of the  API spec. See this link: https://developers.google.com/hangouts/chat/reference/message-formats/cards#card_text_formatting

phillebaba:
@djexp sorry I was  comparing the provider too much to the git providers. From my part this looks good, as the text formatting is documented by google.

I just want to get a second opinion from  @stefanprodan before merging.

phillebaba:
@djexp I just missed that you have not added the new provider to the factory. You need to add a case for the google chat provider in factory.go.
https://github.com/fluxcd/notification-controller/blob/7e9e3f91bd82251b20c32518be37b97a072b4c8f/internal/notifier/factory.go#L43

djexp:
You're right. I've got that locally. Somehow failed to commit it.

I'll take care of that now.

Thanks for being thorough.

phillebaba (approved):
LGTM

djexp:
I'm no expert with Github Actions, but is it normal for actions to be queued for hours like these have been?

stefanprodan:
@djexp GH seems stuck, can you please force push the last commit, maybe it will unblock the actions.

stefanprodan:
> I've included the event timestamp as 'meta-data' within the message. Other implementation have not done this, so I'm happy to remove it for consistency sake if that is preferred.

Please remove it, the timestamp should be available in the chat.

stefanprodan (approved):
LGTM

Thanks @djexp 🏅

Issue #150, opened 2021-03-02
Annotate with the controller-runtime unstructured client 


stefanprodan:
We should drop the dependencies to the GitOps Toolkit API packages and use the controller-runtime [unstructured client](https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/client/unstructured_client.go) to [annotate](https://github.com/fluxcd/notification-controller/blob/main/internal/server/receiver_handlers.go#L342) the GitOps Toolkit objects.

SomtochiAma:
I am picking up this issue.

PR #151, opened 2021-03-03
Uses unstructured client to annotate object

SomtochiAma:
Fixes: #150 

### DIFF
--- a/internal/server/receiver_handlers.go
+++ b/internal/server/receiver_handlers.go
@@ -25,14 +25,15 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"net/http"
 	"net/url"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"strings"
 	"time"
 
-	imagev1 "github.com/fluxcd/image-reflector-controller/api/v1alpha1"
 	"github.com/fluxcd/pkg/apis/meta"
-	sourcev1 "github.com/fluxcd/source-controller/api/v1beta1"
 	"github.com/google/go-github/v32/github"
 	corev1 "k8s.io/api/core/v1"
 	apimeta "k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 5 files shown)


internal/server/receiver_handlers.go
```diff
@@ -344,62 +347,39 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource v1beta1.CrossNam
...
-			source.Annotations = make(map[string]string)
-		}
-		source.Annotations[meta.ReconcileRequestAnnotation] = metav1.Now().String()
-		if err := s.kubeClient.Update(ctx, &source); err != nil {
-			return fmt.Errorf("unable to annotate %s '%s' error: %w", resource.Kind, resourceName, err)
-		}
-	default:
-		return fmt.Errorf("kind '%s not suppored", resource.Kind)
+	apiVersionMap := map[string]string{
+		sourcev1.BucketKind:         "source.toolkit.fluxcd.io/v1beta1",
```
stefanprodan, line 356: We need to get rid of the sourcev1 and imagev1 imports, please hardcode here the Kinds for now.

stefanprodan: We should use the map only if the `APIVersion` is not specified https://github.com/fluxcd/notification-controller/blob/main/api/v1beta1/reference_types.go#L24

config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.4.1
```
stefanprodan, line 7: Please delete `controller-gen` from your GOBIN and rerun make test.

internal/server/receiver_handlers.go
```diff
@@ -344,62 +345,44 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource v1beta1.CrossNam
...
-		if source.Annotations == nil {
-			source.Annotations = make(map[string]string)
-		}
-		source.Annotations[meta.ReconcileRequestAnnotation] = metav1.Now().String()
-		if err := s.kubeClient.Update(ctx, &source); err != nil {
-			return fmt.Errorf("unable to annotate %s '%s' error: %w", resource.Kind, resourceName, err)
-		}
-	default:
-		return fmt.Errorf("kind '%s not suppored", resource.Kind)
+	APIVersionMap := map[string]string{
```
stefanprodan, line 353: Please don't use upper case vars in functions.

internal/server/receiver_handlers.go
```diff
@@ -344,62 +345,44 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource v1beta1.CrossNam
...
+		"Bucket":          "source.toolkit.fluxcd.io/v1beta1",
+		"HelmRepository":  "source.toolkit.fluxcd.io/v1beta1",
+		"GitRepository":   "source.toolkit.fluxcd.io/v1beta1",
+		"ImageRepository": "image.toolkit.fluxcd.io/v1alpha1",
+	}
+
+	APIVersion := APIVersionMap[resource.Kind]
```
stefanprodan, line 360: This should be:

```go
         apiVersion := resource.APIVersion 
	if apiVersion == "" {
		apiVersion = apiVersionMap[resource.Kind]
	}
```

internal/server/receiver_handlers.go
```diff
@@ -357,9 +357,12 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource v1beta1.CrossNam
 		"ImageRepository": "image.toolkit.fluxcd.io/v1alpha1",
 	}
 
-	apiVersion := apiVersionMap[resource.Kind]
-	if resource.APIVersion != "" {
-		apiVersion = resource.APIVersion
+	apiVersion := resource.APIVersion
+	if apiVersion == "" {
+		if apiVersionMap[resource.Kind] == "" {
+			return fmt.Errorf("apiVersion must be specified")
```
stefanprodan, line 363: How can someone tell which kind needs an apiVersion?

stefanprodan (approved):
LGTM

Thanks @SomtochiAma ☕

PR #154, opened 2021-03-05
Fix Azure Devops notifier issues

phillebaba:
This change contains general fixes to the Azure DevOps notifier to fix smaller usability issues.

* Respect schema set by the provider url
* Modify the Azure DevOps notifier to allow compatibility with running behind a proxy.
* Check existing commit statuses to avoid spamming the same status when reconciling.

Signed-off-by: Philip Laine <philip.laine@xenit.se>

### DIFF
--- a/internal/notifier/azure_devops.go
+++ b/internal/notifier/azure_devops.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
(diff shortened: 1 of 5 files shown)


stefanprodan (approved):
LGTM

Thanks @phillebaba 🥇

PR #156, opened 2021-03-15
Use controller-runtime structured logging

stefanprodan:
Use controller-runtime structured logging, required by: https://github.com/fluxcd/flux2/pull/1075


hiddeco (commented):
Wouldn't it be much easier to do something like `logger := s.logger.WithValues(...values)`, and then use this `logger` for as long as the values apply?

hiddeco (approved):

Issue #163, opened 2021-03-17
Different alerts mix up summary


fgunited:
### Issue description

I've created different alerts, listening to GitRepository and HelmRelease. 
They have the same eventSources, the same exclusionList, but different summaries and different providers.
The notifications created by these alert-events will be pushed as expected to the different providers.

**Not working as expected**: But the summary seems to be mixed up. 

Using Notification-Controller v0.9.0, v0.10.0

### Steps to reproduce the issue

Creating three different providers and three alerts, like e.g.:
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: btest-alert-1
  namespace: flux-system
spec:
  summary: "Bugtest Helm Notification"
  providerRef: 
    name: webhook-provider
  eventSeverity: info
  eventSources:
    - kind: "GitRepository"
      name: "*"
      namespace: "flux-system"
    - kind: "HelmRelease"
      name: "*"
      namespace: "dev"
  exclusionList:
    - "HelmChart.*is.not.ready"
    - "i/o.*timeout"
```
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: btest-alert-2
  namespace: flux-system
spec:
  summary: "Another Helm Notification"
  providerRef: 
    name: webhook-provider-2
  eventSeverity: info
  eventSources:
    - kind: "GitRepository"
      name: "*"
      namespace: "flux-system"
    - kind: "HelmRelease"
      name: "*"
      namespace: "dev"
  exclusionList:
    - "HelmChart.*is.not.ready"
    - "i/o.*timeout"	
```
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: btest-alert-3
  namespace: flux-system
spec:
  summary: "Rocketchat Helm Notification"
  providerRef: 
    name: rocketchat-provider
  eventSeverity: info
  eventSources:
    - kind: "GitRepository"
      name: "*"
      namespace: "flux-system"
    - kind: "HelmRelease"
      name: "*"
      namespace: "dev"
  exclusionList:
    - "HelmChart.*is.not.ready"
    - "i/o.*timeout"		
```

### What's the expected result?
Each notification has the correct summary of a defined alert.

### What's the actual result?
[...]

PR #166, opened 2021-03-20
Use deepcopy method for event

SomtochiAma:
Fixes: #163 

Also removes `pkg/recorder` and uses `pkg/runtime/events`


stefanprodan (approved):
LGTM

Thanks @SomtochiAma please rebase with main.

PR #167, opened 2021-03-21
Add rate limiter to event http servers

### LINKED ISSUE #158: HelmRelease errors spam slack
In setting up an alert to send to slack for `HelmRelease` objects at `error` level, it sends a ton of messages.  Here is a simple HelmRelease I applied, with an alert defined for `error`:

```
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: broken-test
  namespace: default
spec:
  chart:
    spec:
      chart: charts/mychart
      sourceRef:
        kind: GitRepository
        name: gitops
        namespace: flux-system
  interval: 10m0s
  values:
    nothing: noone
```
This chart has a required value, so I knew this would trigger an error.  However, slack sent 22 messages within 3 seconds or so, all nearly identical.  We got 11 identical notifications for `Helm install failed` and 11 identical notifications for `reconciliation failed`.

Would it be possible to either thread these messages, or have a setting to limit number of messages within a given time frame? I could image something like this on the AlertSpec:

```
kind: Alert
metadata:
  name: myalert
  namespace: flux-sustem
spec:
  thread: true # For each alert, thread additional messages for the same object type/name
  notifyWindow: 
    numMessages: 1
    timeDelay: 10m0s # Allow only one notification per object per error within the given time frame
```

### LINKED ISSUE #179: Add Prometheus instrumentation to the HTTP servers
We could create a middleware that instruments the events and the webhook receiver servers, this way users can track how many requests have been processed, the number of failures, rate limits, etc.

phillebaba:
Adds a rate limiter to the events http server. The limiters are keyed based on the request.

The event endpoint is keyed based on the involved object, severity of the event, and if present reference. A new severity for a throttled involved object should not be throttled and the previous severity throttling should be reset. This is useful when KC or HC will reconcile a resource multiple times if the reconcile errors. A large amount of duplicate events will be sent during a short amount of time until the back off becomes large enough. This protects NC from spamming providers with the same events.

Fixes #158
Fixes #179


main.go
```diff
@@ -135,14 +137,17 @@ func main() {
 	// +kubebuilder:scaffold:builder
 
 	ctx := ctrl.SetupSignalHandler()
+	store, err := memorystore.New(&memorystore.Config{
+		Interval: 10 * time.Minute,
```
phillebaba, line 141: Make interval configurable with a flag.

internal/server/event_server_test.go
```diff
@@ -0,0 +1,148 @@
...
+			rateLimit: false,
+		},
+		{
+			involvedObject: corev1.ObjectReference{
+				APIVersion: "kustomize.toolkit.fluxcd.io/v1beta1",
+				Kind:       "Kustomization",
+				Name:       "1",
+				Namespace:  "1",
+			},
+			severity:  recorder.EventSeverityInfo,
```
phillebaba, line 92: A new severity should reset any previous rate limits.

internal/server/event_server_test.go
```diff
@@ -0,0 +1,148 @@
+/*
+Copyright 2020 The Flux authors
```
phillebaba, line 2: Should this be 2021?

internal/server/event_server.go
```diff
@@ -69,4 +79,22 @@ func (s *EventServer) ListenAndServe(stopCh <-chan struct{}) {
...
 		s.logger.Info("Event server stopped")
 	}
+
+	return nil
+}
+
+func eventKeyFunc(r *http.Request) (string, error) {
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return "", err
```
phillebaba, line 89: What effect does returning an error here have on the status code and response message?

main.go
```diff
@@ -135,14 +137,17 @@ func main() {
...
+	})
 
 	setupLog.Info("starting event server", "addr", eventsAddr)
 	eventServer := server.NewEventServer(eventsAddr, log, mgr.GetClient())
-	go eventServer.ListenAndServe(ctx.Done())
+	go eventServer.ListenAndServe(ctx.Done(), store)
```
phillebaba, line 146: Either the error should be handled or the function should not return an error.

internal/server/event_server.go
```diff
@@ -69,4 +79,22 @@ func (s *EventServer) ListenAndServe(stopCh <-chan struct{}) {
...
+	}
+	defer r.Body.Close()
+
+	event := &recorder.Event{}
+	err = json.Unmarshal(body, event)
+	if err != nil {
+		return "", err
+	}
+
+	return fmt.Sprintf("event/%s/%s", event.InvolvedObject.String(), event.Severity), nil
```
stefanprodan, line 99: We need to do a checksum of the name, namespace, kind and event message. As it this, means that we drop health checks after apply. Please see my proposal in the linked issues.

stefanprodan (changes requested):

internal/server/event_server.go
stefanprodan: See https://github.com/fluxcd/notification-controller/blob/main/internal/server/event_handlers.go#L44 a 500 error would make the sender retry and we don't want this.

internal/server/event_server_test.go
stefanprodan: Yes

main.go
stefanprodan: What error are we taking about? If the server listener encounters an error it crashes the whole process https://github.com/fluxcd/notification-controller/blob/main/internal/server/receiver_server.go#L59

internal/server/event_server_test.go
phillebaba: Solved when message is included in the checksum

main.go
phillebaba: This is a new error introduced when creating the middleware, but I will solve this by also exiting.

internal/server/event_server.go
phillebaba: This does not solve the problem of a even going from one state to the other as the message will still be the same. One option is to just ignore it and accept it as a limitation.

This test fails but according to me should pass.
https://github.com/fluxcd/notification-controller/blob/7c13a5f58ca0a46fbdc8411346c48788700543ee/internal/server/event_server_test.go#L80

internal/server/event_server.go
phillebaba: A rate limited request will receive the status 429 and retryablehttp client will only retryn on 500 responses, so we should be fine.

internal/server/event_server.go
```diff
@@ -70,3 +83,22 @@ func (s *EventServer) ListenAndServe(stopCh <-chan struct{}) {
...
+
+	event := &events.Event{}
+	err = json.Unmarshal(body, event)
+	if err != nil {
+		return "", err
+	}
+
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(body))
+
+	val := fmt.Sprintf("event/%s/%s/%s/%s", event.InvolvedObject.Name, event.InvolvedObject.Namespace, event.InvolvedObject.Kind, event.Message)
```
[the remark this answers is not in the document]
phillebaba: Yep that is done now, it will append the revision to the key if it is set.

main.go
```diff
@@ -74,6 +79,7 @@ func main() {
 	flag.IntVar(&concurrent, "concurrent", 4, "The number of concurrent notification reconciles.")
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
+	flag.IntVar(&rateLimitInterval, "rate-limit-interval", 5, "Interval in which rate limit has effect, value is given in minutes.")
```
stefanprodan, line 82: Let's make this a DurationVar with `5m` as default like here: https://github.com/fluxcd/kustomize-controller/blob/main/main.go#L79

internal/server/receiver_server.go
```diff
@@ -71,3 +82,10 @@ func (s *ReceiverServer) ListenAndServe(stopCh <-chan struct{}) {
 		s.logger.Info("Receiver server stopped")
 	}
 }
+
+func receiverKeyFunc(r *http.Request) (string, error) {
+	id := url.PathEscape(strings.TrimLeft(r.RequestURI, "/hook/"))
+	val := strings.Join([]string{"receiver", id}, "/")
```
stefanprodan, line 88: We need to checksum the request body if it exists, as it is, we’ll block valid push webhooks for different commits/repos.

stefanprodan (changes requested):

internal/server/receiver_server.go
phillebaba: We are pushing this issue forward as there will be complications with implementing rate limiting to the receiver. Instead we will solve it in a separate PR.

docs/spec/v1beta1/event.md
```diff
@@ -55,3 +55,13 @@ const (
 
 Controller implementations can use the [fluxcd/pkg/runtime/events](https://github.com/fluxcd/pkg/tree/main/runtime/events)
 package to push events to notification-controller API.
+
+## Rate limiting
+
+Events sent to the notification-controller are subject to rate limiting to reduce the amount of duplicate events sent by notification-controller. Events are rate limited based on its `InvolvedObject.Name`, `InvolvedObject.Namespace`, `InvolvedObject.Kind`, and `Message` and if present in the metadata `revision`. The interval of the rate limit is set by default to `5m` but can be configured with the `--rate-limit-interval` option.
```
stefanprodan, line 61:
```suggestion
Events sent to the notification-controller are subject to rate limiting to reduce the amount of duplicate alerts sent by notification-controller. Events are rate limited based on its `InvolvedObject.Name`, `InvolvedObject.Namespace`, `InvolvedObject.Kind`, and `Message` and if present in the metadata `revision`. The interval of the rate limit is set by default to `5m` but can be configured with the `--rate-limit-interval` option.
```

stefanprodan (approved):
LGTM

Thanks @phillebaba 🏅

Issue #133, opened 2021-01-27
Add Webex as an Alert Provider


gdoctor:
I am happy to implement this myself based off the existing providers and open a PR if members of this community are okay with that.

stefanprodan:
I thought Webex is like Zoom, video meetings only. Does it have chat channels like Slack?

gdoctor:
Webex has a full chat client component that was called Webex Teams up until recently. I think they just brought it into the Webex suite and now it is being branded as Webex for messaging; it is essentially a chat client similar to Slack. For reference: https://www.webex.com/team-collaboration.html

phillebaba:
The only issue I see with implementing this feature would be testing and debugging issues.

gdoctor:
I understand your concerns, but Webex has free tier accounts that should have complete access to the messaging components just like MSTeams. I am happy to implement and test and attach all pertinent screenshots.

gdoctor:
I added support for Webex on my end, and I am using the image for my `notification-controller` deployment currently. Before I open a PR here. I want to make sure the flux developers are okay with the content of the message and formatting of it. I initially wanted to use an Adaptive Card like the MSTeams integration, but Webex does not support Adaptive Cards via the Incoming Webhook integration yet so I am left with markdown options. Also, I couldn't get json to look very good on webex.

The screenshot is directly from the Webex client. Let me know what you think @stefanprodan @phillebaba 

![image](https://user-images.githubusercontent.com/5168204/109234946-8e7d6080-7781-11eb-8b77-cf708eea6c66.png)

gdoctor:
This is what the full json looks like. The user scrolls to the right to see the rest of the message.

<img width="1169" alt="Screen Shot 2021-02-25 at 4 06 08 PM" src="https://user-images.githubusercontent.com/5168204/109235944-832b3480-7783-11eb-8646-4b221ef2f3b4.png">

phillebaba:
Just so that I understand the screenshots correctly, will the user see the formatted text or the JSON? I think that even if Webex is not totally identical to the other providers it should use the same format in the messages.

gdoctor:
Sorry my messages were not clear. I provided 2 separate options for what the user could see. The first one is a format that I thought would be good. The second is what the user would see if I instead sent json. I prefer the first option.

gdoctor:
[...]

PR #168, opened 2021-03-22
Add support for Webex as an alert provider

gdoctor:
Fix: #133


gdoctor:
@phillebaba sorry for the delay. Everything should be good now.

stefanprodan (approved):
LGTM

Thanks @gdoctor

PR #169, opened 2021-03-24
Remove filter for progressing events in googlechat

SomtochiAma:
Fixes: #165 

### DIFF
--- a/internal/notifier/google_chat.go
+++ b/internal/notifier/google_chat.go
@@ -91,12 +91,6 @@ func (s *GoogleChat) Post(event events.Event) error {
 		return nil
 	}
 
-	// Skip progressing events. For some reason, these are coming through and then Google Chat seems to replace
-	// the more meaningful message with these which is unhelpful!
-	if event.Reason == "Progressing" {
-		return nil
-	}
-
 	// Header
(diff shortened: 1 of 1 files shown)


SomtochiAma:
@djexp I couldn't see why this was done and it is causing some message to not be sent.

> I've had to filter out 'Progressing' events as these seemed to replace the previous message within Google Chat. I couldn't easily find out why. However the output in chat is very similar to Discord so why this system does not display these messages is a mystery?

Perhaps you could explain further or attach some screenshots?

djexp:
@SomtochiAma The behaviour I was seeing was that Google Chat would seemingly drop all but the latest message that arrived either 'at the same' or within a very very short window. There is no documentation from Google that I could find explaining any such filtering, and there were no errors or status reported through the API to suggest that any such thing was occurring. This just seemed to fit the circumstances.

At the time, using the system I was integration testing with; the Google Chat notifier, (with the filtering) seemed to produce the same output as the Discord and MS Teams implementations; however I didn't have the knowledge with the wider system to investigate further.

I should add that it did not sit right with me to add the filtering code, but without it an incomplete picture was being presented in Google Chat. Clearly if my view on what an 'incomplete picture' is was wrong then this filter should be removed.

stefanprodan (approved):
LGTM

PR #171, opened 2021-03-25
Add observedGeneration to status of alert and receivers

SomtochiAma:

### DIFF
--- a/controllers/alert_controller.go
+++ b/controllers/alert_controller.go
@@ -88,6 +88,8 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 	r.recordReadiness(ctx, alert)
 
+	alert.Status.ObservedGeneration = alert.Generation
+
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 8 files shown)


stefanprodan:
@SomtochiAma we need to also set it like this: https://github.com/fluxcd/kustomize-controller/blob/main/api/v1beta1/kustomization_types.go#L251

controllers/alert_controller.go
```diff
@@ -88,6 +88,8 @@ func (r *AlertReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 
 	r.recordReadiness(ctx, alert)
 
+	alert.Status.ObservedGeneration = alert.Generation
```
stefanprodan, line 91: Hmm what is this supposed to do?

SomtochiAma: Setting it like you stated above?

stefanprodan: What about persistence?

stefanprodan (approved):
LGTM

Thanks @SomtochiAma

PR #174, opened 2021-03-26
fix: use sha1 sum as value for key in bitbucket api requests

### LINKED ISSUE #173: bitbucket max key length
Hi,

Bitbucket api has a max key value length of 40 characters. Currently the value of the key for notification sent to bitbucket is the object kind and name separated by slash i.e `kustomization/service-foo`. If you happen to have long service names and/or combinations with namespace names because of multiple environments sharing the same config repo (`kustomization/namespace-foo-and-service-bar`) you hit the roof which results in 400 error responses from bitbucket.

bergemalm:
Fixes #173 

Suggestion is to use sha1 sum as value.

Signed-off-by: Mikael Bergemalm <mbergemalm@gmail.com>

### DIFF
--- a/internal/notifier/bitbucket.go
+++ b/internal/notifier/bitbucket.go
@@ -85,6 +85,8 @@ func (b Bitbucket) Post(event events.Event) error {
 	}
 
 	name, desc := formatNameAndDescription(event)
+	// key has a limitation of 40 characters in bitbucket api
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

Thanks @bergemalm 🏅

PR #164, opened 2021-03-17
Record suspension metrics

SomtochiAma:
Part of https://github.com/fluxcd/flux2/issues/590
Signed-off-by: Somtochi Onyekwere <somtochionyekwere@gmail.com>

PR #175, opened 2021-03-26
Release v0.11.0

stefanprodan:


CHANGELOG.md
```diff
@@ -2,6 +2,33 @@
 
 All notable changes to this project are documented in this file.
 
+## 0.11.0
+
+**Release date:** 2021-03-26
+
+This is the eleventh MINOR prerelease.
+
+This prerelease comes with support for sending alerts to Webex and
```
hiddeco, line 11:
```suggestion
This prerelease comes with a breaking change to the leader election ID
from `4ae6d3b3.fluxcd.io` to `notification-controller-leader-election`
to be more descriptive. This change should not have an impact on most
installations, as the default replica count is `1`, if you are running
a setup with multiple replicas, it is adviced to scale down before
upgrading.

Support has been added for sending alerts to Webex and
```

stefanprodan: The notification APIs are pure configuration, there is no need to scale down the replicas.

hiddeco: Ack, I think it would still be great to mention breaking changes explicitly as it is good practice, and may help people identify unexpected behavior. So:

```suggestion
This prerelease comes with a breaking change to the leader election ID
from `4ae6d3b3.fluxcd.io` to `notification-controller-leader-election`
to be more descriptive. This change should however not have a direct impact.

Support has been added for sending alerts to Webex and
```

hiddeco (approved):
:rocket:

hiddeco:
Think https://github.com/fluxcd/notification-controller/pull/164 is missing from the changelog.

PR #177, opened 2021-03-29
Add support for signing outgoing webhooks.

bigkevmcd:
This adds a mechanism for a shared secret between the sender and the receiver,
to validate the origin of a webhook when using the generic hook forwarder.

Currently, it's hard-coded to use SHA256 to HMAC the request body.


stefanprodan:
@bigkevmcd can you please add a section to the API spec docs to document this feature?

stefanprodan:
@bigkevmcd any plans to finish this? If not can we close it?

PR #184, opened 2021-04-14
Add self signed cert to provider

### LINKED ISSUE #183: Self Signed Cert on Github Provider
## Description
No option to specify TLS CA self signed cert in Provider Spec.

### Details
We run a GitHub enterprise server that has a self signed CA cert attached to it. When trying to set up git commit status, the notification controller gets the below error.

```
failed to send notification could not list commit statuses: ... : x509: certificate signed by unknown authority
```

Maybe a solution similar to the [ImageRepository](https://toolkit.fluxcd.io/components/image/imagerepositories/) CRD can be implemented in the provider as well. Specifically a value similar to `certSecretRef`?

phillebaba:
Fixes #183 

Adds and optional cert secret to the provider which will be added to the different notifiers http clients. This implementation skips the following notifiers as they are not possible to self host.
* Slack
* Discord
* Teams
* Google Chat


internal/notifier/client.go
```diff
@@ -40,6 +42,9 @@ func postMessage(address, proxy string, payload interface{}, reqOpts ...requestO
 		}
 		httpClient.HTTPClient.Transport = &http.Transport{
 			Proxy: http.ProxyURL(proxyURL),
+			TLSClientConfig: &tls.Config{
+				RootCAs: certPool,
```
stefanprodan, line 46: Is this the right place? Do we want to support CAs only when using a proxy?

config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml
```diff
@@ -50,6 +50,16 @@ spec:
                 description: HTTP/S webhook address of this provider
                 pattern: ^(http|https)://
                 type: string
+              certSecretRef:
```
simongottschlag, line 53: Could it be wise to use caCertSecretRef or something at least containing CA? Easily confused with "client cert".

controllers/provider_controller.go
```diff
@@ -121,7 +122,28 @@ func (r *ProviderReconciler) validate(ctx context.Context, provider v1beta1.Prov
...
+
+		caFile, ok := secret.Data["caFile"]
+		if !ok {
+			return fmt.Errorf("no caFile found in secret %q", provider.Spec.CertSecretRef.Name)
+		}
+
+		certPool = x509.NewCertPool()
+		ok = certPool.AppendCertsFromPEM(caFile)
+		if !ok {
+			return fmt.Errorf("could not append to cert pool")
```
simongottschlag, line 142: Will this error ever show up when debugging? Could it be caused by someone uploading a cert with the wrong format, encoding etc? Maybe give brief description about what could be causing it.

internal/notifier/sentry.go
```diff
@@ -28,9 +32,14 @@ type Sentry struct {
 }
 
 // NewSentry creates a Sentry client from the provided Data Source Name (DSN)
-func NewSentry(dsn string) (*Sentry, error) {
+func NewSentry(certPool *x509.CertPool, dsn string) (*Sentry, error) {
 	client, err := sentry.NewClient(sentry.ClientOptions{
 		Dsn: dsn,
+		HTTPTransport: &http.Transport{
```
simongottschlag, line 38: No if statement needed like the others?

config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml
phillebaba: Follows standard set by other controllers.
https://toolkit.fluxcd.io/components/image/imagerepositories/

controllers/provider_controller.go
phillebaba: I dont see how we could give more details as no error is returned.

controllers/provider_controller.go
```diff
@@ -121,7 +122,28 @@ func (r *ProviderReconciler) validate(ctx context.Context, provider v1beta1.Prov
...
+		var secret corev1.Secret
+		secretName := types.NamespacedName{Namespace: provider.Namespace, Name: provider.Spec.CertSecretRef.Name}
+
+		if err := r.Get(ctx, secretName, &secret); err != nil {
+			return fmt.Errorf("failed to read secret, error: %w", err)
+		}
+
+		caFile, ok := secret.Data["caFile"]
+		if !ok {
+			return fmt.Errorf("no caFile found in secret %q", provider.Spec.CertSecretRef.Name)
```
stefanprodan, line 136: Please don’t use %q, we log in json format and this will cause escape characters.

controllers/provider_controller.go
stefanprodan: We should append the secret name, e.g. "could not append to cert pool: invalid CA found in %s”

stefanprodan (approved):
LGTM

Thanks @phillebaba 🎖️

Issue #190, opened 2021-05-03
Azure EventHub integration


simongottschlag:
We are building an integration for the notification-controller to work with Azure EventHub: https://azure.microsoft.com/en-us/services/event-hubs/

PR #191, opened 2021-05-03
Add support for Azure EventHub provider

NissesSenap:
Solves #190


api/v1beta1/provider_types.go
```diff
@@ -28,7 +28,7 @@ const (
 // ProviderSpec defines the desired state of Provider
 type ProviderSpec struct {
 	// Type of provider
-	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry
+	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry;eventhub;
```
phillebaba, line 31: Might be better to call it `azureeventhub` as there may be other event hubs we dont know about which could confuse people.

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+	"context"
+	"encoding/json"
+	"fmt"
+	"time"
+
+	eventhub "github.com/Azure/azure-event-hubs-go/v3"
+	"github.com/fluxcd/pkg/runtime/events"
+)
+
+type EventHub struct {
```
simongottschlag, line 26: Add comment for struct

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+	Hub *eventhub.Hub
+}
+
+func NewEventHub(endpointURL string) (*EventHub, error) {
```
simongottschlag, line 30: Add comment for func

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+	hub, err := eventhub.NewHubFromConnectionString(endpointURL)
+	if err != nil {
+		return nil, err
+	}
+	return &EventHub{
+		Hub: hub,
+	}, nil
+}
+
+// Post EventHub msg
```
simongottschlag, line 40: Add more info

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+func (e *EventHub) Post(event events.Event) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	eventBytes, err := json.Marshal(event)
+	if err != nil {
+		return fmt.Errorf("Unable to marshall event: %w", err)
+	}
+
+	//err := e.Hub.Send(ctx, eventhub.NewEventFromString("From, String"))
```
simongottschlag, line 50: Can be removed?

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+	err = e.Hub.Send(ctx, eventhub.NewEvent(eventBytes))
+	if err != nil {
+		return fmt.Errorf("Failed to send msg: %w", err)
+	}
+
+	err = e.Hub.Close(context.Background())
```
simongottschlag, line 56: Could it be possible for Close() to hang? Perhaps should use the context with timeout.

internal/notifier/eventhub.go
```diff
@@ -0,0 +1,61 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &EventHub{
+		Hub: hub,
+	}, nil
+}
+
+// Post EventHub msg
+func (e *EventHub) Post(event events.Event) error {
```
simongottschlag, line 41: I see that some of the notifiers are using pointers and some are not. Is there a reason for one or the other?

simongottschlag (commented):
Left a few comments 👍😊

internal/notifier/eventhub.go
NissesSenap: What do you mean? I can't find any pointer with event events.Event. It wouldn't work.

internal/notifier/eventhub.go
NissesSenap: Haha I have created above but for some stupid reason not used it...

NissesSenap: I have created it above but for some reason I don't use it... Fixed that now

internal/notifier/eventhub.go
simongottschlag: Above is using `func (e *Eventhub) ...`. 

The other notifiers are using a mix of `*` and not.

My question is (and this may be obvious and I just don't fully understand it) why are the different notifiers doing this differently and what should be used?

docs/spec/v1beta1/provider.md
```diff
@@ -46,6 +46,7 @@ Notification providers:
 * Google Chat
 * Webex
 * Sentry
+* EventHub
```
simongottschlag, line 49: Azure

internal/notifier/azure_eventhub.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &AzureEventHub{
+		Hub: hub,
+	}, nil
+}
+
+// Post all notification-controller messages to EventHub
+func (e *AzureEventHub) Post(event events.Event) error {
```
simongottschlag, line 43: Doesn't really matter, but maybe change the variable from `e` to `a`

simongottschlag (commented):
Minor tweaks

internal/notifier/eventhub.go
NissesSenap: We can take this offline

stefanprodan:
@NissesSenap please run `make test` before pushing, this will fix go mod, code formatting, etc.

internal/notifier/azure_eventhub.go
```diff
@@ -0,0 +1,117 @@
...
+	"github.com/fluxcd/pkg/runtime/events"
+)
+
+// AzureEventHub holds the eventhub client
+type AzureEventHub struct {
+	Hub *eventhub.Hub
+}
+
+// NewAzureEventHub creates a eventhub client
+func NewAzureEventHub(endpointURL, token, eventhubNamespace string) (*AzureEventHub, error) {
```
phillebaba, line 33: Nit

eventhubNamespace -> eventHubNamespace

internal/notifier/azure_eventhub.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	return &AzureEventHub{
+		Hub: hub,
+	}, nil
+}
+
+// Post all notification-controller messages to EventHub
+func (e *AzureEventHub) Post(event events.Event) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
```
phillebaba, line 57: You should add a check if the event is an update event as is done here.
https://github.com/fluxcd/notification-controller/blob/15d98c9afca937d01c64c7753df4615045f2fc65/internal/notifier/google_chat.go#L90

phillebaba (approved):

docs/spec/v1beta1/provider.md
```diff
@@ -222,3 +223,105 @@ SECRET_NAME=tls-certs
...
+
+If you want to make a easy test assuming that you have setup a Azure Enterprise application and you called it event-hub you can follow most of the bellow commands.
+You will need to provide the client_secret that you got when generating the Azure Enterprise Application.
+
+```shell
+export AZURE_CLIENT=$(az ad app list --filter "startswith(displayName,'event-hub')" --query '[].appId' |jq -r '.[0]')
+export AZURE_SECRET='secret-client-secret-generated-at-creation'
+export AZURE_TENANT=$(az account show -o tsv --query tenantId)
+
+curl -X GET --data 'grant_type=client_credentials' --data "client_id=$AZURE_CLIENT" --data "client_secret=$AZURE_SECRET" --data 'resource=https://eventhubs.azure.net' -H 'Content-Type: application/x-www-form-urlencoded' https://login.microsoftonline.com/$AZURE_TENANT/oauth2/token |jq .access_token
```
simongottschlag, line 276: https://docs.microsoft.com/en-us/cli/azure/account?view=azure-cli-latest#az_account_get_access_token

simongottschlag (commented):
nit

docs/spec/v1beta1/provider.md
NissesSenap: This will use your account instead of the application account. Risk of giving to much access, I would recomend to use this way.

phillebaba (approved):
LGTM :rose:

PR #193, opened 2021-05-06
Add note about expoosing receiver to the internet

### LINKED ISSUE #180: Add rate limiter for receiver
Exposing the receiver to the internet may result in an accidental dos of the api server, either through a bug in the other end making a request to the receiver or a bad actor figuring out the receiver endpoint.

The rate limiter is implemented in the events server in #167 but was not implemented for the receiver. The reason for this was because there is a high risk that too many requests would be rate limited as it is hard to differentiate requests for different commit references.

phillebaba:
Adds a small note about the risks of exposing the receiver to the public internet and how to mitigate them.

Fixes #180

### DIFF
--- a/docs/spec/v1beta1/receiver.md
+++ b/docs/spec/v1beta1/receiver.md
@@ -63,6 +63,19 @@ type ReceiverStatus struct {
 }
 ```
 
+## Public ingress considerations
+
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @phillebaba

PR #196, opened 2021-05-10
Redact token from error log

### LINKED ISSUE #160: Redact tokens from logs
It would be nice if we could redact tokens from logs.

phillebaba:
Some errors like those from net/http will include the token value. We need to make sure that we redact the token value before logging as to not expose the value.

Fixes #160


stefanprodan (approved):
LGTM

Thanks @phillebaba

PR #207, opened 2021-06-04
Add cert pool to Slack provider requests

### LINKED ISSUE #203: Support MatterMost
We use mattermost (slack clone) in our enterprise and after much debugging using the slack provider i discovered the code does not contain the recently added logic to handle self-signed certificates.

Can this please be added in a future release?

phillebaba:
This change just brings the Slack provider inline with other providers which could be used with self hosted instances.

Fixes #203


stefanprodan (approved):
LGTM

Thanks @phillebaba 💯

Issue #200, opened 2021-05-14
Slack provider for a private channel doesn't work


messiahUA:
I have an incoming webhook configured for a specific private channel and as far as I know slack has made a change at some point that it's forbidden to specify the channel name in this case, so it's impossible to use such webhook because "channel" is mandatory https://github.com/fluxcd/notification-controller/blob/337b5ab624512818ccbba6e35fecc750a004b770/internal/notifier/slack.go#L69

When I'm testing the webhook url using curl it works perfectly fine without specifying the channel, but if I specify it I always get "channel_not_found". And notification controller is completely silent about this.

phillebaba:
Is there a way that we can differentiate between the webhooks that require a channel and webhooks that dont?

messiahUA:
I don’t think there is a way to do that based only on the webhook url. And as far as I know webhooks should have a default channel configured, so it should be left for the flux user to specify the channel or not and thus it should be optional.

If I’m wrong and there is an option that channel may be indeed mandatory (on the slack side) then controller should just log an error returned from the slack.

SomtochiAma:
I will take a look at this issue

SomtochiAma:
Hey @messiahUA,
Specifying the name of the channel still works for me, events get posted to the slack channel.
I agree with you though we should consider the name of channel not being required. But specifying it in this case still works for me.
I have been unable to reproduce this. Or is there something I am missing?

SomtochiAma:
Making a curl request(with data `'{"text":"Hello, World!", "channel":"test-private"}'`) also works.
Did you structure your curl request differently?

messiahUA:
I've tried once again with the same webhook and it works. Even tried to different private channels. I'm confused, not sure what's going on here and what changed... Maybe I made a mistake, but not sure how it is possible because I did same simple curl request. Anyways, can't explain it right now.

stefanprodan:
This could’ve been a transient issue on the Slack side.

SomtochiAma:
I noticed that specifying any random string as the channel name still works(not adding the field still works too).
It seems the channel field is not actually required for slack.
@stefanprodan wdyt? Should we make the `channel` field optional for slack?

stefanprodan:
There was no default channel required in Slack UI, if this has changed then yes we could remove our validation.

PR #208, opened 2021-06-07
Make slack channel optional

SomtochiAma:
## Changes
- Makes slack channel optional
- Only redacts token from logs if it is present (to prevent distorting the error message)
- Checks status code of HTTP request and returns an error if not successful

Fixes: #200 

### DIFF
--- a/internal/notifier/client.go
+++ b/internal/notifier/client.go
@@ -21,6 +21,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


internal/notifier/client.go
```diff
@@ -87,9 +88,18 @@ func postMessage(address, proxy string, certPool *x509.CertPool, payload interfa
 	for _, o := range reqOpts {
 		o(req)
 	}
-	if _, err := httpClient.Do(req); err != nil {
+	resp, err := httpClient.Do(req)
+	if err != nil {
 		return fmt.Errorf("failed to execute request: %w", err)
 	}
 
+	if resp.StatusCode != http.StatusOK {
```
stefanprodan, line 96: What about HTTP 202?

SomtochiAma: Will add that in

internal/notifier/client.go
```diff
@@ -87,9 +88,18 @@ func postMessage(address, proxy string, certPool *x509.CertPool, payload interfa
...
+	if err != nil {
 		return fmt.Errorf("failed to execute request: %w", err)
 	}
 
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusCreated {
+		b, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("unable to read response body, %s", err)
+		}
+		return fmt.Errorf("request not successful, %s", string(b))
```
stefanprodan, line 101:
```suggestion
		return fmt.Errorf("request failed with status code %: %s", resp.StatusCode, string(b))
```

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🌷

Issue #194, opened 2021-05-06
Adding cluster name (or something else) to Github commit status context


gmaiztegi:
We run a few clusters with Flux and are moving towards Flux 2 now, and so far the experience is being awesome.

We have seen an issue in this notification controller, and it's that when having multiple clusters posting the commit status for the same resource names (such as `kustomization/flux-system`), the statuses get overwritten, as they use the `<resource type>/<name>` as context.

Therefore, it would be awesome if we could add some kind of prefix or suffix to this context in order to distinguish the status posted by the different clusters.

And once again, Flux 2 is great. Thanks for your hard work.

phillebaba:
I agree that we should have a second look at this. When we first implemented the commit status feature a lot was unknown. Currently I am solving this by adding the environment or cluster name to the Kustomization resource. This will cause the statuses to be unique between different clusters.

One solution may be to expose some sort of format configuration for the provider. A risk is that users would hit the character limit for a status message. Most of the providers have one but they are not documented so they are found by trial and error. Do you have any suggestion for how this configuration should be implemented in the provider or alert resource @gmaiztegi?

gmaiztegi:
I'm not sure on wether it should be on the provide or the alert. It seems that this is a Github specific issue regarding the commit status, although the question might be wether it might be useful to have different alerts with different identifiers, all of them pointing to the Github provider. It's not our use case, nor I can think of any right now, but this seems more flexible.

gdoctor:
On a related note, adding some cluster metadata to a notification would be extremely valuable. I am also running flux on 20+ clusters now, and we are unable to use the notification controller effectively since we cannot tell where the notification originated without having a separate chat room per notification-controller instance (we are going to have 50 or more clusters so not very reasonable).

stefanprodan:
@gdoctor you can add the cluster name to the alert message with `spec.summary`, docs here: https://fluxcd.io/docs/components/notification/alert/

bforchhammer:
Hi, I've also run into the issue described here, and adding the "summary" to an "alert" unfortunately does not seem to do anything in the case of github status messages; github status seem to only contain the `kind`/`name` of the referenced object and the `event reason` as defined in `formatNameAndDescription` here: https://github.com/fluxcd/notification-controller/blob/c4377ceb51f0ed969aa3fac1c151189bfa18d7fc/internal/notifier/util.go#L47

jrauschenbusch:
[...]

PR #209, opened 2021-06-07
Custom Git Commit Status Prefix 

clarkdoan:
Resolves: #194 

### Problem
When setting a git commit status the controller only uses the  `event.InvolvedObject.Kind` and  `event.InvolvedObject.Name` to generate the status' context. This does not allow multiple clusters to update the commit status if they use the same Kustomization name. 

### Proposed Solution
The event that is passed to the `formatNameAndDescription` function contains a `Metadata` field. 

The `AlertSpec` contains a `Summary` field which is used as a `Short description of the impact and affected cluster.`

The event handler copies the value of the `Summary` field from the `AlertSpec` to the `Metadata` field of the event. 

This PR checks the value of the `Metadata` field to see if there is a `summary` key. If the key exists then the summary value will be prepended to the context of the commit status. This will allow the user to set the prefix of the commit status.

### DIFF
--- a/api/v1beta1/alert_types.go
+++ b/api/v1beta1/alert_types.go
@@ -47,6 +47,7 @@ type AlertSpec struct {
 	ExclusionList []string `json:"exclusionList,omitempty"`
 
 	// Short description of the impact and affected cluster.
+	// For commit status updates, this field is used to prefix the status message.
 	// +optional
 	Summary string `json:"summary,omitempty"`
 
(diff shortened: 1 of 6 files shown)


stefanprodan (commented):
Please change the `Summary` description in code and run `make test` then commit the generated CRDs. Append this to the field description: "For commit status updates, this field is used to prefix the status message.".

phillebaba:
We need to do some testing with this feature regarding different provider length constraints and format constraints, as this will allow users to inject arbitrary text into the status ID.

stefanprodan:
@phillebaba I think we should trim the summary to a small number of characters, as you said we need to determine which number that is.

phillebaba:
Github has a limit of 140 last time I hit this issue but is not documented. Gitlab has something similar while Azure DevOps has a much larger limit. The issue is that none of these limits are documented.

phillebaba:
I would actually prefer to hold off on this PR and instead implement a more generic solution as proposed here.
https://github.com/fluxcd/flux2/discussions/1529

PR #223, opened 2021-07-26
providers/sentry: add environment support

BeryJu:
use channel configuration for sentry environment to re-use the same DSN for multiple clusters

created this as a separate PR to not clash with the other sentry PR

### DIFF
--- a/docs/spec/v1beta1/provider.md
+++ b/docs/spec/v1beta1/provider.md
@@ -226,7 +226,20 @@ kubectl create secret generic $SECRET_NAME \
 
 ### Sentry
 
-The sentry provider uses the `channel` field to specify which environment the messages are sent for.
+The sentry provider uses the `channel` field to specify which environment the messages are sent for:
+
+```yaml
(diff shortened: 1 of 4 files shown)


BeryJu:
sorry for the flood of PRs, I have one other PR prepared after this one is merged

stefanprodan:
Can you please add a note [in here](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta1/provider.md#notifications) to how `Channel` is used for sentry? Without docs no one would guess this.

BeryJu:
Of course, you reckon I should just create a new category somewhere along line 217?

docs/spec/v1beta1/provider.md
```diff
@@ -224,6 +224,10 @@ kubectl create secret generic $SECRET_NAME \
   --from-file=caFile=ca.crt
 ```
 
+### Sentry
+
+The sentry provider uses the `channel` field to specify which environment the messages are sent for.
```
stefanprodan, line 229: Can you please add here an yaml example that includes `channel `? Thanks.

PR #224, opened 2021-07-26
providers/sentry: send traces

BeryJu:
This is the last sentry PR, I promise 😄 

This PR adds support to send Sentry Traces instead of just errors. This means that normal info events are sent as traces and errors are sent as errors. 

This functionality can be disabled by setting `eventSeverity` in the related `Alert` object to `error`


stefanprodan (approved):
LGTM

Thanks @BeryJu 🎖️

PR #231, opened 2021-08-08
feat: set default observedGeneration to -1 on CRDs

makkes:
This sets the `status.observedGeneration` field to -1 by default. This
is a follow-up to https://github.com/fluxcd/helm-controller/pull/294,
porting the same code to the notification-controller so that all Flux
2 controllers work the same way in this regard.

Signed-off-by: Max Jonas Werner <mail@makk.es>


stefanprodan (approved):
LGTM 

Thanks @makkes

PR #232, opened 2021-08-16
Add Telegram notification

### LINKED ISSUE #222: Add support for Telegram notifications
Hey, it would be great if you could add support for Telegram notifications.

You just have make a POST or GET request to specific URL with a few parameters.
The follwing docs show the basics:
https://core.telegram.org/bots/api#making-requests
And for the message one need to use the following method:
https://core.telegram.org/bots/api#sendmessage
The user would only need to supply the URL and the chat_id.
As message formatting, markdown and html is possible.

If you need any further information, feel free to ask.
Thank you in advance :)

SomtochiAma:
Closes #222


stefanprodan:
@SomtochiAma please rebase with main branch and solve the conflicts for CI to run.

hiddeco (commented):
What is the rationale behind using `shoutrrr` for Telegram only, instead of e.g. using a Telegram API library, or using it for other providers as well?

stefanprodan:
I think we should look into switching to `shoutrrr` for the providers that are supported by it. For example people were asking for SlackBot support as an alternative to the classic webhook.

internal/notifier/telegram.go
```diff
@@ -36,10 +36,29 @@ func (t *Telegram) Post(event events.Event) error {
...
 	%s`, emoji, strings.ToLower(event.InvolvedObject.Kind),
 		event.InvolvedObject.Name, event.InvolvedObject.Namespace, event.Message)
-	url := fmt.Sprintf("telegram://%s@telegram?channels=%s", t.Token, t.Channel)
-	err := shoutrrr.Send(url, message)
+	url := fmt.Sprintf("telegram://%s@telegram?channels=%s&parseMode=markDownv2", t.Token, t.Channel)
+	err := shoutrrr.Send(url, escapeString(message))
 	return err
 }
+
+func escapeString(str string) string {
```
stefanprodan, line 48: Is this something that shoutrrr requires or it's telegram? Can you please link to some docs around the escape logic?

SomtochiAma: Telegram requires it when using Markdown. Docs: (https://core.telegram.org/bots/api#formatting-options)[https://core.telegram.org/bots/api#formatting-options]

stefanprodan: Ok please add a comment to this function including the link.

internal/notifier/telegram.go
```diff
@@ -0,0 +1,68 @@
...
+		event.InvolvedObject.Name, event.InvolvedObject.Namespace, event.Message)
+	url := fmt.Sprintf("telegram://%s@telegram?channels=%s&parseMode=markDownv2", t.Token, t.Channel)
+	err := shoutrrr.Send(url, escapeString(message))
+	return err
+}
+
+// The telegram API requires that some special characters are escaped
+// in the message string. Docs: https://core.telegram.org/bots/api#formatting-options.
+// The function only handles a subset of the characters because the only special
+// characters in kubernetes names and namespaces are '-' and '.'.
```
stefanprodan, line 51: But `event.Message` can contain any character, for example a kubectl apply error.

SomtochiAma: I will update the function

stefanprodan: Please test this with a Kustomization that fails to apply. I see that we don't post the revision but we should, check how alerts look for Slack: https://fluxcd.io/img/slack-error-alert.png

stefanprodan: We have to add the event metadata to the message body as separate field in the markdown, maybe a list?

SomtochiAma: This is what the message looks like. With the metadata added

Screenshot 2021-08-18 at 12 35 06 PM

stefanprodan (approved):
LGTM 

Thanks @SomtochiAma please squash your commits

L3o-pold:
Seems that's not working for `HelmRelease` upgrade notification. Unfortunatly I haven't the error from the Telegram return body.
```json
{"level":"info","ts":"2021-09-22T12:19:25.795Z","logger":"event-server","msg":"Dispatching event: Helm upgrade succeeded","reconciler kind":"HelmRelease","name":"ingress-nginx","namespace":"flux-system"}
{"level":"error","ts":"2021-09-22T12:19:25.816Z","logger":"event-server","msg":"failed to send notification","reconciler kind":"HelmRelease","name":"ingress-nginx","namespace":"flux-system","error":"failed to send notification to \"-****\", response status code 400 Bad Request"}
```

PR #233, opened 2021-08-18
Matrix Notification

### LINKED ISSUE #229: Add support for notifying into matrix
Matrix (matrix.org) is an open source federated chat is more and more used as a replacement for Slack.
Matrix has a very simple API that is alread implemented by services such as gitea.

The typical parameters for a matrix webhook are:

* Homeserver URL
* Room ID (where to post)
* Access token (authentication)
* Event type (optional, can be m.notice or m.text)

SomtochiAma:
Closes #229 


dholbach:
Not sure if this would make things more complicated, but the matrix provider could use `shoutrrr` too: https://containrrr.dev/shoutrrr/v0.5/services/overview/

SomtochiAma:
>  matrix provider could use shoutrrr too

I missed that. I will update it so that it used shoutrr instead

SomtochiAma:
@dholbach I am currently having an issue using the current version of  `shoutrrr` with `matrix` so I am looking into it. But, I have opened another PR that makes it easier to add notification providers supported by Shoutrr

PR #235, opened 2021-08-25
Use Shoutrr type for telegram notification

SomtochiAma:

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -137,6 +137,7 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 
 			webhook := provider.Spec.Address
 			token := ""
+			port := ""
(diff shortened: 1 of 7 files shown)


internal/server/event_handlers.go
```diff
@@ -157,6 +158,10 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 				if t, ok := secret.Data["token"]; ok {
 					token = string(t)
 				}
+
+				if p, ok := secret.Data["port"]; ok {
```
stefanprodan, line 162: Why is the port a "secret"? The URL should contain the port if it's not 80 or 443.

SomtochiAma: I will take the port out

dholbach:
Best rename the file to `shoutrrr.go` (with 3x "r").

PR #243, opened 2021-09-09
Controller refactoring for various reconcilers

### LINKED ISSUE #251: Update notification-controller to new unified standards
- Part of https://github.com/fluxcd/flux2/issues/1601

This issue is to track rolling out the changes to Notification Controller.

This controller is sufficiently complex to make exemplary use of pkg/runtime, specifically it's primary purpose which is move conditions logic to a reusable package. Once this is updated, it should help steer the way for a wider group to roll these standards out to the other controllers.

SomtochiAma:
Supersedes [scottrigby/notification-controller#1](https://github.com/scottrigby/notification-controller/pull/1/files)

Co-authored-by: Piaras Hoban <piaras@weave.works>
Co-authored-by: Hidde Beydals <hiddeco@users.noreply.github.com>
Co-authored-by: souleb <bah.soule@gmail.com>

Closes #251
Ref https://github.com/fluxcd/flux2/issues/1601


api/v1beta1/provider_types.go
```diff
@@ -22,7 +22,8 @@ import (
 )
 
 const (
-	ProviderKind string = "Provider"
+	ProviderKind     string = "Provider"
+	ProviderIndexKey string = ".metadata.provider"
```
hiddeco, line 26: Not sure if this value belongs in the API package, as based on the implementation it seems to be part of the application logic, and would have little value to consumers.

controllers/receiver_controller.go
```diff
@@ -55,54 +62,115 @@ type ReceiverReconciler struct {
...
+	token, err := r.token(ctx, obj)
+	if err != nil {
+		conditions.MarkFalse(obj, meta.ReadyCondition, v1beta1.TokenNotFoundReason, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	receiverURL := fmt.Sprintf("/hook/%s", sha256sum(token+obj.Name+obj.Namespace))
+
+	// Nothing has changed so return early
+	if conditions.IsReady(obj) && obj.Status.URL == receiverURL && obj.Status.ObservedGeneration == obj.Generation {
```
hiddeco, line 158: Checking a previous recorded condition makes the object state machine like, which it shouldn't be. Instead, it should simply rely on comparing the spec of the object to the current state, and make adjustments if they do not add up, and then push the condition.

controllers/alert_controller.go
```diff
@@ -21,146 +21,180 @@ import (
...
+			readyCondition := conditions.Get(alert, meta.ReadyCondition)
+			switch readyCondition.Status {
+			case metav1.ConditionFalse:
+				// As we are no longer reconciling and the end-state is not ready, the reconciliation has stalled
+				conditions.MarkStalled(alert, readyCondition.Reason, readyCondition.Message)
+			case metav1.ConditionTrue:
+				// As we are no longer reconciling and the end-state is ready, the reconciliation is no longer stalled
+				conditions.Delete(alert, meta.StalledCondition)
+			}
+		}
```
hiddeco, line 137: You may want to have a look and see if the above works in all scenarios, as Michael figured it didn't in some: https://github.com/fluxcd/pkg/pull/144

hiddeco:
@SomtochiAma you requested a review but I do not see any changes since my previous review round. Did you forget to push?

controllers/alert_controller.go
```diff
@@ -21,146 +21,184 @@ import (
...
 
-func (r *AlertReconciler) validate(ctx context.Context, alert v1beta1.Alert) error {
-	var provider v1beta1.Provider
-	providerName := types.NamespacedName{Namespace: alert.Namespace, Name: alert.Spec.ProviderRef.Name}
-	if err := r.Get(ctx, providerName, &provider); err != nil {
-		return fmt.Errorf("failed to get provider %s, error: %w", providerName.String(), err)
-	}
+func (r *AlertReconciler) reconcile(ctx context.Context, alert *v1beta1.Alert) (ctrl.Result, error) {
+	// Mark the resource as under reconciliation
+	conditions.MarkReconciling(alert, meta.ProgressingReason, "")
```
hiddeco, line 156: Fail to observe where this is unset.

SomtochiAma: In the defer func in the main reconcile func

controllers/alert_controller.go
```diff
@@ -21,146 +21,184 @@ import (
...
-		return fmt.Errorf("provider %s is not ready", providerName.String())
+	// validate alert spec and provider
+	if err := r.validate(ctx, alert); err != nil {
+		conditions.MarkFalse(alert, meta.ReadyCondition, meta.FailedReason, err.Error())
+		return ctrl.Result{}, err
 	}
 
-	return nil
+	conditions.MarkTrue(alert, meta.ReadyCondition, meta.SucceededReason, v1beta1.InitializedReason)
+	ctrl.LoggerFrom(ctx).Info("Alert initialised")
```
hiddeco, line 165:
```suggestion
	ctrl.LoggerFrom(ctx).Info("Alert initialized")
```

to match the reason.

controllers/alert_controller.go
```diff
@@ -21,146 +21,184 @@ import (
...
+
+func (r *AlertReconciler) SetupWithManagerAndOptions(mgr ctrl.Manager, opts AlertReconcilerOptions) error {
+	if err := mgr.GetFieldIndexer().IndexField(context.TODO(), &v1beta1.Alert{}, ProviderIndexKey,
+		func(o client.Object) []string {
+			alert := o.(*v1beta1.Alert)
+			return []string{
+				fmt.Sprintf("%s/%s", alert.GetNamespace(), alert.Spec.ProviderRef.Name),
+			}
+		}); err != nil {
+
```
hiddeco, line 74: Missing return.

controllers/alert_controller.go
```diff
@@ -21,146 +21,184 @@ import (
...
-		return fmt.Errorf("failed to get provider %s, error: %w", providerName.String(), err)
-	}
+func (r *AlertReconciler) reconcile(ctx context.Context, alert *v1beta1.Alert) (ctrl.Result, error) {
+	// Mark the resource as under reconciliation
+	conditions.MarkReconciling(alert, meta.ProgressingReason, "")
 
-	if !apimeta.IsStatusConditionTrue(provider.Status.Conditions, meta.ReadyCondition) {
-		return fmt.Errorf("provider %s is not ready", providerName.String())
+	// validate alert spec and provider
+	if err := r.validate(ctx, alert); err != nil {
```
hiddeco, line 159: Given this validates part of the provider reference, which are being watched. I think you may be able to filter the error out in case the reference does not exist (and not requeue). Once the provider reference is created again, a reconcile will be requested again for the object.

controllers/provider_controller.go
```diff
@@ -44,61 +45,107 @@ import (
...
-	return ctrl.NewControllerManagedBy(mgr).
-		For(&v1beta1.Provider{}).
-		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicates.ReconcileRequestedPredicate{})).
-		Complete(r)
+func (r *ProviderReconciler) reconcile(ctx context.Context, obj *v1beta1.Provider) (ctrl.Result, error) {
+	// Mark the resource as under reconciliation
+	conditions.MarkReconciling(obj, meta.ProgressingReason, "")
+
+	// validate provider spec and credentials
+	if err := r.validate(ctx, obj); err != nil {
```
hiddeco, line 137: As this validates both the spec, and references itself, I think it would be wise to filter the error. As invalid spec errors do not require a requeue, given we do not expect to recover without a generation change of the object.

SomtochiAma: I think that in a case where the secret(`provider.Spec.SecretRef`) is missing, the secret could be created and this could cause a recover without a generation change of the object.

controllers/alert_controller.go
SomtochiAma: I don't know if we should log the not found errors since we are filtering it. Nothing shows up in the logs just in the condition of the alert resource

hiddeco: I would log the error I think.

controllers/provider_controller.go
hiddeco: You are absolutely right!

hiddeco (approved):
This looks good to me, but would like another pair of :eyes: more familiar with this controller to have another look. Thanks @SomtochiAma :1st_place_medal:

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🏅 

PS. We need to add tests for the receiver reconciler and make sure update to `spec` doesn't change the generated URL.

PR #252, opened 2021-10-06
Add opsgenie provider and additional metadata

Alan01252:
This pull request adds opsgenie as a provider and allows extra metadata to be configured against an alert.
This is useful when you're running multiple clusters and want to distinguish where the alert has come from ( for example dev/staging/prod environments )

Opsgenie has been added as a receiver and this has been tested locally. 
Opsgenie makes use of the extra metadata in order to route requests to appropriate slack channels etc.


api/v1beta1/alert_types.go
```diff
@@ -54,6 +54,11 @@ type AlertSpec struct {
 	// Defaults to false.
 	// +optional
 	Suspend bool `json:"suspend,omitempty"`
+
+	// Extra metadata to be added to the alert
+	// +kubebuilder:validation:Optional
+	// +optional
+	ExtraMetadata map[string]string `json:"extraMetadata,omitempty"`
```
stefanprodan, line 61: Please remove this field, the `summary` can be used to set the cluster name, region, etc.

stefanprodan (changes requested):

api/v1beta1/alert_types.go
Alan01252: By adding this map you get a lot more flexibility around being able to process the alert once it reached the provider. 

The summary as a string means more complex parsing logic at the provider end. There are many tags that may/may not make sense in a summary field which adding a map would allow.

It also better fits in with the Prometheus "labels" paradigm, which many people running Kubernetes would be used to. I'm thinking an "alertmanager" provider could also make heavy use of this in the same way I am using this with the Prometheus provider too.

stefanprodan: Once we defines the set of annotations for the Event, you could define routing rules based on e.g. `GIT_URL` and check if the URL matches some application repo. Besides the Event annotations, we could include the Alert annotations too, so instead of adding some info in the Alert `.spec.extraMetadata` you would add it to `.metadata.annotations`.

stefanprodan: For now please map the `summary` field to the opsgenie details.

Alan01252:
As discussed I've removed the extraMetadata functionality and confirmed summary field appears in opsgenie if added to the Alert manifest.

image

docs/spec/v1beta1/provider.md
```diff
@@ -9,7 +9,7 @@ Spec:
 ```go
 type ProviderSpec struct {
 	// Type of provider
-	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;
+	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;opsgenie
```
stefanprodan, line 12: Please add opsgenie in the `Notification providers` list down below.

Alan01252:
Interesting this is failing. I was debating whether I need to check-in the output of `make manifest` but this causes a rather larger diff than I was expecting so I assumed I was missing something. I could just manually add opsgenie to the card which would resolve this..

stefanprodan:
You need to run `make generate api-docs` and commit the changes.

Alan01252:
> You need to run `make generate api-docs` and commit the changes.

There are no changes to the API docs when running that command, the build was failing because of a diff in the CRD docs. If I run `make manifests` locally I get a much larger diff that expected.

Alan01252:
image

internal/notifier/opsgenie.go
```diff
@@ -0,0 +1,77 @@
...
+	_, err := url.ParseRequestURI(hookURL)
+	if err != nil {
+		return nil, fmt.Errorf("invalid Opsgenie hook URL %s", hookURL)
+	}
+
+	return &Opsgenie{
+		URL:      hookURL,
+		ProxyURL: proxyURL,
+		CertPool: certPool,
+		ApiKey:   token,
```
stefanprodan, line 52: Is the token required? If so we should error out if it's an empty string. If this is required, I think we should add an Opsgenie section to the API docs and show people how to create the secret and which keys to set.

stefanprodan: Thank you!

PR #258, opened 2021-10-11
Add alertmanager provider

Alan01252:
This commit adds the alertmanager provider. The provider adds some
generic labels based on the event which should be enough to configure
appropraite routes within alertmanager.

The alert is annotated with the message by default and optionally by the
summary field given in the event.


Alan01252:
Hmm not sure why DCO is hanging now.

Alan01252:
> Hmm not sure why DCO is hanging now.

tried another push seems to be still be busted...

PR #282, opened 2021-11-01
Use conditions helper in reconciler (kstatus compat)

SomtochiAma:
Ref: https://github.com/fluxcd/flux2/issues/1601

Co-authored-by: Piaras Hoban <piaras@weave.works>
Co-authored-by: Hidde Beydals <hiddeco@users.noreply.github.com>
Co-authored-by: souleb <bah.soule@gmail.com>


controllers/alert_controller.go
```diff
@@ -21,145 +21,185 @@ import (
...
+	defer func() {
+		patchOpts := []patch.Option{
+			patch.WithOwnedConditions{
+				Conditions: []string{
+					meta.ReadyCondition,
+					meta.ReconcilingCondition,
+					meta.StalledCondition,
+				},
+			},
+			patch.WithStatusObservedGeneration{},
```
hiddeco, line 123: This is conditionally added on L129, and should be removed here to ensure the observed generation is only bumped once the resource has ended up in a "finite state" for that generation (by either successfully having reconciled to a `Ready==True` / `Stalled==True` /  `Reconciling==False` state).

controllers/alert_controller.go
```diff
@@ -21,145 +21,185 @@ import (
...
-	}
+func (r *AlertReconciler) reconcile(ctx context.Context, alert *v1beta1.Alert) (ctrl.Result, error) {
+	// Mark the resource as under reconciliation
+	conditions.MarkReconciling(alert, meta.ProgressingReason, "")
 
-	if !apimeta.IsStatusConditionTrue(provider.Status.Conditions, meta.ReadyCondition) {
-		return fmt.Errorf("provider %s is not ready", providerName.String())
+	// validate alert spec and provider
+	if err := r.validate(ctx, alert); err != nil {
+		conditions.MarkFalse(alert, meta.ReadyCondition, meta.FailedReason, err.Error())
```
hiddeco, line 159: Given we know in detail what failed here, it would probably be more meaningful to make this a `ValidationFailed` reason. 

The Reason is not part of the API and can thus be arbitrary, but if it can be recorded by multiple reconcilers, you may still want to factor it out to a defined type anyway.

controllers/alert_controller.go
```diff
@@ -21,145 +21,185 @@ import (
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	return nil
+	conditions.MarkTrue(alert, meta.ReadyCondition, meta.SucceededReason, v1beta1.InitializedReason)
```
hiddeco, line 163: Wondering if there is something more meaningful to share here than re-using a generic Reason as a Message.

controllers/alert_controller.go
```diff
@@ -21,145 +21,185 @@ import (
...
-func (r *AlertReconciler) recordSuspension(ctx context.Context, alert v1beta1.Alert) {
-	if r.MetricsRecorder == nil {
-		return
+func (r *AlertReconciler) validate(ctx context.Context, alert *v1beta1.Alert) error {
+	provider := &v1beta1.Provider{}
+	providerName := types.NamespacedName{Namespace: alert.Namespace, Name: alert.Spec.ProviderRef.Name}
+	if err := r.Get(ctx, providerName, provider); err != nil {
+		// log not found errors since they get filtered out
+		ctrl.LoggerFrom(ctx).Error(err, "failed to get provider %s, error: %w", providerName.String())
+		return fmt.Errorf("failed to get provider '%s', error: %w", providerName.String(), err)
```
hiddeco, line 175:
```suggestion
		// log not found errors since they get filtered out
		ctrl.LoggerFrom(ctx).Error(err, "failed to get provider %s", providerName.String())
		return fmt.Errorf("failed to get provider '%s': %w", providerName.String(), err)
```

Using `, error:` is kind of redundant since Go has added support for proper error wrapping. In addition, the logger already takes `err` as an error, and the second argument sugars this error with an informational message.

controllers/provider_controller.go
```diff
@@ -44,61 +45,108 @@ import (
...
+	defer func() {
+		patchOpts := []patch.Option{
+			patch.WithOwnedConditions{
+				Conditions: []string{
+					meta.ReadyCondition,
+					meta.ReconcilingCondition,
+					meta.StalledCondition,
+				},
+			},
+			patch.WithStatusObservedGeneration{},
```
hiddeco, line 102: This is conditionally added on L108, and should be removed here to ensure the observed generation is only bumped once the resource has ended up in a "finite state" for that generation (by either successfully having reconciled to a `Ready==True` / `Stalled==True` /  `Reconciling==False` state).

controllers/provider_controller.go
```diff
@@ -44,61 +45,108 @@ import (
...
-		For(&v1beta1.Provider{}).
-		WithEventFilter(predicate.Or(predicate.GenerationChangedPredicate{}, predicates.ReconcileRequestedPredicate{})).
-		Complete(r)
+func (r *ProviderReconciler) reconcile(ctx context.Context, obj *v1beta1.Provider) (ctrl.Result, error) {
+	// Mark the resource as under reconciliation
+	conditions.MarkReconciling(obj, meta.ProgressingReason, "")
+
+	// validate provider spec and credentials
+	if err := r.validate(ctx, obj); err != nil {
+		conditions.MarkFalse(obj, meta.ReadyCondition, meta.FailedReason, err.Error())
```
hiddeco, line 139: Same as for the Alert reconciler.

controllers/provider_controller.go
```diff
@@ -44,61 +45,108 @@ import (
...
+		return ctrl.Result{}, err
+	}
+
+	conditions.MarkTrue(obj, meta.ReadyCondition, meta.SucceededReason, v1beta1.InitializedReason)
```
hiddeco, line 143: Same as for the Alert reconciler.

controllers/receiver_controller.go
```diff
@@ -55,53 +62,116 @@ type ReceiverReconciler struct {
...
+		// Patch the object, ignoring conflicts on the conditions owned by this controller
+		patchOpts := []patch.Option{
+			patch.WithOwnedConditions{
+				Conditions: []string{
+					meta.ReadyCondition,
+					meta.ReconcilingCondition,
+					meta.StalledCondition,
+				},
+			},
+			patch.WithStatusObservedGeneration{},
```
hiddeco, line 97: Same as for the other reconcilers.

controllers/receiver_controller.go
```diff
@@ -55,53 +62,116 @@ type ReceiverReconciler struct {
...
+		conditions.MarkFalse(obj, meta.ReadyCondition, v1beta1.TokenNotFoundReason, err.Error())
+		return ctrl.Result{}, err
+	}
+
+	receiverURL := fmt.Sprintf("/hook/%s", sha256sum(token+obj.Name+obj.Namespace))
+
+	// Nothing has changed so return early
+	if obj.Status.URL == receiverURL && obj.Status.ObservedGeneration == obj.Generation {
+		return ctrl.Result{}, nil
+	}
```
hiddeco, line 161: Is this check idempotent? E.g. if L150 fails due to a transient error and records `Ready == False`, wouldn't this prevent it from recovering to `Ready==True`?

controllers/alert_controller.go
SomtochiAma: Not sure. The major event/check that has happened is that the provider that is referenced by the alert is present and ready

controllers/receiver_controller.go
SomtochiAma: Good catch. It does prevent it from recovering. I will remove the check.

hiddeco (approved):
This looks good to me, thanks all :bow: 

It would be great if eventually the test suite can also be ported, as I think the current one is quite minimal.

Issue #286, opened 2021-11-17
Notification controller crashes when trying to send alerts to Alertmanager  


zdkhan:
When the notification controller tries to send an alert to Alertmanager it throws an error and crashes the pod. 

Flux version is v0.23.0
Current version of notification controller running in cluster is v0.18.1

Below is the yaml used to setup the provider and alerts 

```
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Provider
metadata:
  name: alertmanager
  namespace: flux-system
spec:
  type: alertmanager
  address: http://<myurl>/api/v2/alerts/
---
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: errors
  namespace: flux-system
spec:
  providerRef:
    name: alertmanager
  eventSeverity: error
  eventSources:
    - kind: GitRepository
      name: '*'
    - kind: Kustomization
      name: '*'
    - kind: HelmRelease
      name: '*'
    - kind: ImageRepository
      name: '*'
    - kind: ImagePolicy
      name: '*'
    - kind: ImageUpdateAutomation
      name: '*'  
```

The logs show an event being dispatched but then errors out.
```
`{"level":"info","ts":"2021-11-17T17:04:45.649Z","logger":"event-server","msg":"Dispatching event: version list argument cannot be empty","reconciler kind":"ImagePolicy","name":"my-app","namespace":"flux-system"}`

panic: assignment to entry in nil map

goroutine 353 [running]:
github.com/fluxcd/notification-controller/internal/notifier.(*Alertmanager).Post(0xc000a506c0, 0xc000620230, 0xb, 0xc000620240, 0xb, 0xc000046258, 0x18, 0xc0005ac0c0, 0x24, 0xc0004fc140, ...)
	/workspace/internal/notifier/alertmanager.go:69 +0x20e
github.com/fluxcd/notification-controller/internal/server.(*EventServer).handleEvent.func1.1(0x0, 0x0, 0xc000231950, 0xc0004dc2a0, 0x1f17e20, 0xc000a506c0, 0xc000620230, 0xb, 0xc000620240, 0xb, ...)
	/workspace/internal/server/event_handlers.go:225 +0x75
created by github.com/fluxcd/notification-controller/internal/server.(*EventServer).handleEvent.func1
	/workspace/internal/server/event_handlers.go:224 +0xdb8
```

Is there something wrong with my configuration sending alerts to Alertmanager?

[...]

PR #287, opened 2021-11-18
Fix nil dereference err in AlertManager

SomtochiAma:
Fix: #286 


stefanprodan (approved):
LGTM

Thanks @SomtochiAma ☕

PR #291, opened 2021-11-23
Feature: Add option to pass additional headers via secretRef to the generic notification provider

boxcee:
The generic notification provider should have a little more configuration options.

We have a use-case where we want to kick-off Github workflows based on deployments. A feature to add additional headers would allow us to do this. Without additional headers we are not able to authenticate with Github though, since it requires an `Authorization` header with the personal access token set.

### DIFF
--- a/docs/spec/v1beta1/provider.md
+++ b/docs/spec/v1beta1/provider.md
@@ -163,6 +163,28 @@ The body of the request looks like this:
 
 The `involvedObject` key contains the object that triggered the event.
 
+```yaml
+apiVersion: notification.toolkit.fluxcd.io/v1beta1
+kind: Provider
+metadata:
+  name: generic
+  namespace: default
+spec:
+  type: generic
+  address: https://api.github.com/repos/owner/repo/dispatches
+  secretRef: 
+    name: generic-secret
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: generic-secret
+  namespace: default
+data:
+  # 'Authorization: token\nX-Forwarded-Proto: https'
+  headers: QXV0aG9yaXphdGlvbjogdG9rZW4KWC1Gb3J3YXJkZWQtUHJvdG86IGh0dHBz
+```
+
 ### Self-signed certificates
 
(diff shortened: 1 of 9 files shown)


yebyen:
Hi @boxcee ! Thanks for your contribution here,

Can you please update the PR description with some explanation about what is added here? When you are ready to have it reviewed, can you please make sure the DCO bot is satisfied, and that you have squashed and/or `--sign-off`ed your commits?

boxcee:
@yebyen done 👍🏾

modalkonform:
@yebyen 
Is there any chance this change will get merged soon?

docs/spec/v1beta1/provider.md
```diff
@@ -163,6 +163,28 @@ The body of the request looks like this:
...
+  address: https://api.github.com/repos/owner/repo/dispatches
+  secretRef: 
+    name: generic-secret
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: generic-secret
+  namespace: default
+data:
```
stefanprodan, line 183: Please use `stringData` e.g.
```yaml
stringData:
  headers: |
     Authorization: token
     X-Forwarded-Proto: https
```

Also explain how the `headers` are used.

stefanprodan (changes requested):

kingdonb:
Can you provide a bit more detail about the example of use for documentation? I understand the sample and the change, I'm just unsure about how the GitHub side of the configuration would look like, and how Flux users should discover this feature.

The example added in the doc isn't given any context or description in the current revision of the PR. That would make this finished, afaict 👍 @boxcee the branch has merge conflicts right now and so, looks like it must be rebased in order to merge.

stefanprodan:
Superseded by #317

alekspog:
@boxcee hi, how do you trigger Github workflows with /dispatches URL with generic notification provider? It seems that Github repository_dispatch trigger requires a mandatory `type_event` field https://docs.github.com/en/rest/reference/repos#create-a-repository-dispatch-event, but I don't see how to pass custom fields with this generic provider.

PR #294, opened 2021-11-29
Refactor tests

SomtochiAma:
This pull request removes ginkgo from the test. It also adds tests for receivers handler.

New tests added:
`TestReceiverHandler` and `TestEventHandler`(refactored)


controllers/testdata/crd.yaml
```diff
@@ -0,0 +1,284 @@
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
stefanprodan, line 3: Please remove this file, we get this into bin.

go.mod
```diff
@@ -13,14 +13,15 @@ require (
 	github.com/fluxcd/notification-controller/api v0.19.0
 	github.com/fluxcd/pkg/apis/meta v0.11.0-rc.1
 	github.com/fluxcd/pkg/runtime v0.13.0-rc.5
+	github.com/fluxcd/pkg/ssa v0.3.1
+	github.com/fluxcd/source-controller/api v0.19.0
```
stefanprodan, line 17: We can’t make NC depend on any other Flux API. If we need to operate on sources we ca use unstructured objects.

stefanprodan (changes requested):

controllers/testdata/repo.yaml
```diff
@@ -0,0 +1,11 @@
...
+apiVersion: source.toolkit.fluxcd.io/v1beta1
+kind: GitRepository
+metadata:
+  name: podinfo
+  namespace: "%[1]s"
+spec:
+  interval: 1m
+  url: https://github.com/stefanprodan/podinfo
+  ref:
+    branch: main
```
stefanprodan, line 11: Please use here a `semver: 6.0.x`, there is no main brach in podinfo repo.

stefanprodan (commented):
@SomtochiAma you please write in the PR description what this does, which new tests were added and for what. Also please squash the commits.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🌷

PR #298, opened 2022-01-05
Add timestamp as label for alertmanager

Alan01252:
Without the timestamp there is not enough entropy in the hash for
alertmanager to recongise that this is a new alert for some cases.

Adding the timestamp ensures that a new hash for the alert is created
each time and therefore no updates get missed.


Alan01252:
This allows the alertmanager administrator to determine how to group alerts. 

As an example if you choose not to group by timestamp you can send alerts to slack which look like.

image

However if you want to then group by timestamp and get all the alerts

image

stefanprodan (approved):
LGTM

Thanks @Alan01252 🏅

PR #319, opened 2022-01-27
Allow disabling cross-namespace event sources

stefanprodan:
On multi-tenant clusters, platform admins can disable cross-namespace event sources with the `--no-cross-namespace-refs=true` flag. When this flag is set, alerts can only refer to event sources in the same namespace as the alert object, preventing tenants from subscribing to anther tenant's events.

When the controller drops an event due to `--no-cross-namespace-refs=true` it logs:

```
alert 'tenant1/slack' can't process event from 'HelmRelease/tenant2/app', cross-namespace references have been blocked
```

Part of: https://github.com/fluxcd/flux2/issues/2337


internal/server/event_handlers.go
```diff
@@ -95,6 +95,15 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 				if source.Namespace == "" {
 					source.Namespace = alert.Namespace
 				}
+
+				if s.noCrossNamespaceRefs && source.Namespace != alert.Namespace {
```
SomtochiAma, line 99: I don't know if it might be better to check the namespace of the event's `involvedObject` or check after matching sources have been found(L109).
Sending one event produces a log for every source in the alert that is cross-namespaced.

An alternative might be to have this check before the alert is initialized so the alert isn't ready once there are cross-namespace sources and `--no-cross-namespace-refs` is set.

stefanprodan: If an alert contains same -s and cross-ns we forward only those in same ns.

SomtochiAma: Okay, that means we don't need to check when the alert is initialized.
But I think there should be something in the logs that says that we aren't forwarding this event because cross-namespace access has been denied.
Right now, This is the logs
```
{"level":"error","ts":"2022-01-28T07:52:48.216+0100","logger":"event-server","msg":"access denied to cross-namespace source","error":"can't access 'GitRepository/flux-system/*', cross-namespace references have been blocked"}
{"level":"error","ts":"2022-01-28T07:52:48.216+0100","logger":"event-server","msg":"access denied to cross-namespace source","error":"can't access 'Kustomization/flux-system/*', cross-namespace references have been blocked"}
```

and I am not sure if it is really clear

stefanprodan: I've moved the check to involved object, thanks!

stefanprodan: @SomtochiAma I've added the alert name and namespace to the error log, should be more clear now, please check.

SomtochiAma (approved):

PR #324, opened 2022-02-08
discord: Remove username and channel requirements

### LINKED ISSUE #323: discord provider properties
the discord provider does require the username to be set: 
https://github.com/fluxcd/notification-controller/blob/main/internal/notifier/discord.go#L51
The username is optional by discord and not allowing it to be empty doesn't allow to use the ReportingController for the username. 
https://github.com/fluxcd/notification-controller/blob/main/internal/notifier/discord.go#L78
It would be good to make username optional.

The current discord implementation is using the slack compatible webhook. This does not support channel. 
https://discord.com/developers/docs/resources/webhook#execute-slackcompatible-webhook
The discord native webhook also does not support channel, hence removing or at least making channel optional would preferred.

vchrisb:
Fixes #323 

Changes: 
* username optional
* channel optional and remove it from payload

### DIFF
--- a/internal/notifier/discord.go
+++ b/internal/notifier/discord.go
@@ -17,7 +17,6 @@ limitations under the License.
 package notifier
 
 import (
-	"errors"
 	"fmt"
 	"net/url"
 	"path"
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @vchrisb

Issue #275, opened 2021-10-19
Feature Request: Selector-like Config Options for Alert CRD Based on Label


shin-go:
Per the current documentation on the [Alert CRD](https://fluxcd.io/docs/components/notification/alert/), when configuring multiple `EventSources` you must manually enumerate each CRD type and name. Since manually configuring things is something few want to do for very long or at scale, introducing a selector-like option for Alerts could be useful - particularly if it can be based off Labels on the specified `EventSource` kind.

My particular use case for this is that we have multiple dev teams who own multiple microservices. We have each microservice reconciling via a Kustomization specifically for that service (ex.: configrepo/dev-us-west-2/microfoo/, configrepo/dev-us-west-2/microbar/). In order for the wildcard pattern to work, we would need each miroservice subdir in the config repo to be prefixed with a team identifier; or move the subdirs around such that all microservices owned by a particular team is in a subdir for that team (e.g.: `configrepo/dev-us-west-2/foobar/micro[foo|bar]`). While the latter approach *could* work, ownership or dissolution of the team could change down the road.

In this case, relying on Labels on the Kustomization, etc. CRDs to assert ownership wouldn't require much change and is maintainable into the future. A mock of such configuration might look like:

```
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: team-foobar-kustomizations
  namespace: default
spec:
  providerRef: 
    name: dev-msteams
  eventSeverity: error
  eventSources:
    - kind: Kustomization
      namespace: default
      selector:
        kind: [name|label]
        value: 'owner: foobar' # or literal/wildcard as docs currently stand if kind: name
```

stefanprodan:
I propose we add new optional field to the `EventSource` named `matchLabels` of type `map[string]string`. The selector will be taken into consideration only if `name` is set to `*`. This would not break the current API, and will allow filtering of objects of a particular Kind within a namespace based on labels.

This selects all GitRepos and Kustomizations in the `default` namespace:

```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: all
  namespace: default
spec:
  providerRef:
    name: slack
  eventSeverity: info
  eventSources:
    - kind: GitRepository
      name: "*"
    - kind: Kustomization
      name: "*"
```

This selects all GitRepos and Kustomizations in the `default` namespace that have the `team: dev` and `env: staging` labels:

```yaml
[...]

PR #325, opened 2022-02-09
Add LabelSelectors for Event Sources

SomtochiAma:
Fix: #275


controllers/event_handling_test.go
```diff
@@ -163,74 +179,86 @@ func TestEventHandler(t *testing.T) {
...
-					e.InvolvedObject.Name = "test"
-					e.InvolvedObject.Namespace = namespace
-					e.Message = "test"
-					return e
-				},
-				forwarded: true,
+			forwarded: true,
+		},
+		{
+			name: "forwards events the label matches",
```
stefanprodan, line 236: Can you please add a test for "drop events when labels don't match"

internal/server/event_handlers.go
```diff
@@ -96,12 +99,40 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 					source.Namespace = alert.Namespace
 				}
 
-				if (source.Name == "*" || event.InvolvedObject.Name == source.Name) &&
-					event.InvolvedObject.Namespace == source.Namespace &&
+				if event.InvolvedObject.Namespace == source.Namespace &&
```
stefanprodan, line 102: Let's extract this logic to a function called `eventMatchesAlert`

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🏅

PR #330, opened 2022-02-16
Switch to controller-runtime metadata client

SomtochiAma:
Follow-up: #325 


SomtochiAma:
Yes we can, but we still have to provide the group version and kind

stefanprodan:
Ok let’s use the metadata client for all operations.

stefanprodan:
@SomtochiAma can you please add tests for receiver annotations? We only test the reconciliation but not the webhook handler.

stefanprodan (approved):
LGTM

Thanks @SomtochiAma 🌷

PR #331, opened 2022-02-17
Allow the proxy address to specified in the Kubernetes Secret from Alert `spec.secretRef`

fsequeira1:
Signed-off-by: Filipe Sequeira <filipe@weave.works>

### DIFF
--- a/controllers/provider_controller.go
+++ b/controllers/provider_controller.go
@@ -148,6 +148,7 @@ func (r *ProviderReconciler) reconcile(ctx context.Context, obj *v1beta1.Provide
 
 func (r *ProviderReconciler) validate(ctx context.Context, provider *v1beta1.Provider) error {
 	address := provider.Spec.Address
+	proxy := provider.Spec.Proxy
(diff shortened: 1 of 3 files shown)


controllers/provider_controller.go
```diff
@@ -162,6 +163,10 @@ func (r *ProviderReconciler) validate(ctx context.Context, provider *v1beta1.Pro
 			address = string(a)
 		}
 
+		if p, ok := secret.Data["proxy"]; ok {
+			token = string(p)
```
stefanprodan, line 167:
```suggestion
			proxy = string(p)
```

stefanprodan (changes requested):

stefanprodan:
I understand that the proxy setting can be overwritten, but users need to read our source code to figure this out. Please update  the docs and remove the proxy from the example YAML.

stefanprodan (approved):
LGTM

Thanks @fsequeira1 🏅

PR #342, opened 2022-03-14
add fsgroup for securityContext

kingdonb:
Ref: https://github.com/fluxcd/flux2/issues/2537

In https://github.com/fluxcd/source-controller/issues/284#issuecomment-774703153 it seems that leaving `fsGroup` unset can result in this error.

### DIFF
--- a/config/manager/deployment.yaml
+++ b/config/manager/deployment.yaml
@@ -19,6 +19,10 @@ spec:
         prometheus.io/port: "8080"
     spec:
       terminationGracePeriodSeconds: 10
+      securityContext:
+        # Required for AWS IAM Role bindings
+        # https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html
+        fsGroup: 1337
       containers:
         - name: manager
           image: fluxcd/notification-controller


config/manager/deployment.yaml
```diff
@@ -19,6 +19,8 @@ spec:
         prometheus.io/port: "8080"
     spec:
       terminationGracePeriodSeconds: 10
+      securityContext:
+        fsGroup: 1337
```
hiddeco, line 23:
```suggestion
      securityContext:
        # Required for AWS IAM Role bindings
        # https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html
        fsGroup: 1337
```

kingdonb: Does either Notification Controller or Helm Controller actually use AWS IAM role bindings though?

I didn't add this comment because I was unsure of the reason why this might be needed here.

kingdonb: (I accepted the change so you can get this merged and it will conform to the rest, regardless I'd like to understand the answer.)

We had someone on Slack who was actively experiencing this issue but I was not able to get them to confirm this fixes it.

kingdonb: Maybe the answer was, they bound some annotations to AWS IAM roles to Flux service accounts and didn't include it in their report. I would guess the presence of the role binding is what causes the fsGroup setting to be needed, (not the actual utilization of the AWS IAM role) but I would just be guessing at that.

hiddeco: What is interesting is that the latest EKS docs mention the following:

> Providing this security context is not required for 1.19 or later clusters.
> &mdash; https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts-technical-overview.html#Pod-configuration

I think we should create an issue to get to the bottom of this, as it looks like it can be deprecated in the future.

hiddeco:
@kingdonb can you squash please?

hiddeco (approved):

kingdonb:
So there is still https://github.com/fluxcd/flux2/issues/2537 which we can ping for follow-up.

PR #349, opened 2022-03-23
Update the webex notification provider

ahothan:
I have tested this against the latest webex API portal and verified I could get notifications in the targeted webex space.
Requires 2 additional parameters to make it work: a room ID and a Webex bot access token (both can be retrieved using the 
webex developer web site after registering).
I have also enhanced the formatted markdown.

Screen Shot 2022-03-23 at 11 32 38 AM

### DIFF
--- a/fluxcd.code-workspace
+++ b/fluxcd.code-workspace
@@ -1,0 +1,10 @@
+{
+    "folders": [
+        {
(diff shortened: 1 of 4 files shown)


ahothan:
will resubmit new PR

PR #350, opened 2022-03-23
Filter event metadata based on the object group

stefanprodan:
Filter events metadata based on the involved object group.


pjbgf (approved):
LGTM

hiddeco (approved):

PR #352, opened 2022-03-23
Update the webex notification provider and markdown

ahothan:
Signed-off-by: ahothan <ahothan@cisco.com>

I have tested this against the latest webex API portal and verified I could get notifications in the targeted webex space.
Requires 2 additional parameters to make it work: a room ID and a Webex bot access token (both can be retrieved using the
webex developer web site after registering).
I have also enhanced the formatted markdown.
Example of webex space output:

Screen Shot 2022-03-23 at 11 32 38 AM

### DIFF
--- a/internal/notifier/webex.go
+++ b/internal/notifier/webex.go
@@ -23,23 +23,70 @@ import (
 	"strings"
 
 	"github.com/fluxcd/pkg/runtime/events"
+	"github.com/hashicorp/go-retryablehttp"
 )
+
+// Example of provider manifest for webex:
+//
+// apiVersion: notification.toolkit.fluxcd.io/v1beta1
+// kind: Provider
+// metadata:
+//   name: webex
+//   namespace: flux-system
+// spec:
+//   type: webex
+//   address: https://webexapis.com/v1/messages
+//   channel: <webexSpaceRoomID>
+//  secretRef:
+//    name: webex-bot-access-token
+// ---
+// apiVersion: v1
+// kind: Secret
+// metadata:
+//   name: webex-bot-access-token
+//   namespace: flux-system
+// data:
+//   # bot access token - must be base64 encoded
+//   # echo -n <token> | base64
+//   token: <webexBotAccessTokenBase64>
+//
+// General steps on how to hook up Flux notifications to a Webex space:
+// From the Webex App UI:
+// - create a Webex space where you want notifications to be sent
+// - add the bot email address to the Webex space (see next section)
+//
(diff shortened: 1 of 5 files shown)


kingdonb:
Thank you for this, if you are looking to add an example webex notification provider, the file where provider-specific examples are drawn from is: `spec/v1beta1/provider.md`

The example would land in the fluxcd docs under here:
https://fluxcd.io/docs/components/notification/provider/#example

(The current supplementary docs from each controller repo are pulled into the fluxcd/website repo on each deploy)

internal/notifier/webex.go
```diff
@@ -23,23 +23,70 @@ import (
 	"strings"
 
 	"github.com/fluxcd/pkg/runtime/events"
+	"github.com/hashicorp/go-retryablehttp"
 )
 
+// Example of provider manifest for webex:
```
stefanprodan, line 29: Please move the example to the API docs right under this section https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta1/provider.md#slack-app

stefanprodan (changes requested):

internal/notifier/webex.go
ahothan: done

ahothan:
checking if this markdown format would be better:
<Alert summary in bold>
source: <Kind>/<namespace>/<name>
<message>
| <key>=<value> other kv pairs in manifests (all but summary)

stefanprodan:
@ahothan can you please post a print screen of how it looks now?

ahothan:
@stefanprodan 
I wanted the summary to be first row as it would show where the notification.
Depending on whether the alerts have a summary or not, the notification will use one or another layout.
The comment for the CreateMarkdown function describes how the markdown looks like based on summary present or not, 
Here is a snapshot of a mock up (not real notification so dont pay attention to the exact severity)

In this snapshot, the Alert summary was set to "cc-k8s cluster notification" for the first 2,
thelast 2 is without a summary field.

Screen Shot 2022-03-28 at 7 37 54 AM


Please provide any suggestion...

ahothan:
not sure what to make of this error in the "kind" job:

Run kubectl -n notification-system get providers -oyaml
error: the server doesn't have a resource type "providers"
Error: Process completed with exit code 1.

ahothan:
Snapshot from new flux bootstrap with latest version:

Screen Shot 2022-03-28 at 9 41 36 AM

stefanprodan (approved):
LGTM 

Thanks @ahothan 🥇

PR #361, opened 2022-04-08
Return err on invalid UTF-8 character in token

SomtochiAma:
Closes: #360 


stefanprodan (approved):
LGTM 

Thanks @SomtochiAma 🏅

PR #366, opened 2022-04-20
Check for duplicate commit status in Bitbucket

SomtochiAma:
This pull request checks if a similar commit status has been sent to the BitBucket API to avoid spamming. 
This feature is already implemented for the other git providers.

### DIFF
--- a/internal/notifier/bitbucket.go
+++ b/internal/notifier/bitbucket.go
@@ -19,6 +19,7 @@ package notifier
 import (
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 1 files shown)


stefanprodan (commented):
@SomtochiAma can you please confirmed that this was tested with a BitBucket repo and a health check that was previously failing was marked as successful when it recovers?

SomtochiAma:
@stefanprodan yes it was tested

stefanprodan (approved):
LGTM

Thanks @SomtochiAma

PR #385, opened 2022-06-18
Add context to Post function

phillebaba:
The change adds a context to the Post function. Most notifiers do not require a context as they just use HTTP requests, but those that import an SDK tend to do. By passing a context in the Post function new contributions do not have to guess on which context to use and how long the timeout should be.

This change also fixes some context issues, adding context to Gitlab and fixing Azure DevOps context.

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -63,7 +63,7 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 
 		cleanupMetadata(event)
 
-		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		ctx, cancel := context.WithTimeout(r.Context(), 15*time.Second)
 		defer cancel()
 
 		var allAlerts v1beta1.AlertList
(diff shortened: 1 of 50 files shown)


internal/server/event_handlers.go
```diff
@@ -62,7 +62,7 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 
 		cleanupMetadata(event)
 
-		ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+		ctx, cancel := context.WithTimeout(r.Context(), 15*time.Second)
```
phillebaba, line 65: After reading some a bit more on this I am questioning if this is the right choice. My initial thought was that this context would only be cancelled when the HTTP server is shutdown, but it can also be cancelled by the client. It seems like that type of behavior is not really what we want as NC should just attempt to complete the post.

stefanprodan: If the client cancels the context then it should also cancel the event posting, the reason for this is that the clients are the Flux controller which do have retry/backoff and they are responsible for retying the notification.

stefanprodan (approved):
LGTM

Thanks @phillebaba 🏅

PR #401, opened 2022-08-10
Remove code for parsing github payload

### LINKED ISSUE #393: Unable to parse `package` event from GitHub webhook
Error message I get when the webhook comes in.

`unable to parse GitHub payload, err: json: cannot unmarshal object into Go struct field PackageVersion.package.package_version.body of type string`.

Trying to use this with the image watcher.

`v0.22.2`

somtochiama:
Closes #393 

The go-github library doesn't parse the `package` event probably due to some inconsistencies in what Github sends and differences in their API documentation.
Since we don't use the Go struct returned, this code has now been removed.

This pull request also adds tests for the receiver handler.

### DIFF
--- a/internal/server/receiver_handler_test.go
+++ b/internal/server/receiver_handler_test.go
@@ -43,17 +43,6 @@ func Test_validate(t *testing.T) {
 		header    string
 	}
 
-	status := v1beta1.ReceiverStatus{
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @somtochiama 🏅

PR #404, opened 2022-08-15
flag added to Disable http scheme for providers

gunishmatta:
Hi,
Please review and suggest changes.

Thanks


main.go
```diff
@@ -82,6 +83,7 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", 5*time.Minute, "Interval in which rate limit has effect.")
+	flag.BoolVar(&httpSchemeDisabled, "http-scheme-enabled", false, "Enable Http Scheme When true, the flag would not allow the use of the http scheme across all controller-level objects.")
```
pjbgf, line 86: We could name it in a way that it portrays this is not the secure thing to do:
```suggestion
	flag.BoolVar(&insecureNoTLS, "insecure-no-tls", false, "Enable the use of the HTTP scheme (no TLS) across all controller-level objects. This is not recommended for production environments. ")
```

internal/server/event_server.go
```diff
@@ -44,15 +44,17 @@ type EventServer struct {
 	logger               logr.Logger
 	kubeClient           client.Client
 	noCrossNamespaceRefs bool
+	httpSchemeDisabled   bool
```
pjbgf, line 47:
```suggestion
	supportHttpScheme   bool
```

internal/server/event_handlers.go
```diff
@@ -243,6 +243,13 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 				continue
 			}
 
+			if s.httpSchemeDisabled && strings.Contains(webhook, "http://") {
```
pjbgf, line 246: Let's use `url.Parse` and validate for `.Scheme == "http"`.

pjbgf (changes requested):

gunishmatta:
@pjbgf I have done the suggested changes.
Please review

main.go
```diff
@@ -82,6 +83,7 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", 5*time.Minute, "Interval in which rate limit has effect.")
+	flag.BoolVar(&insecureNoTLS, "insecure-no-tls", false, "Enable the use of HTTP Scheme (no TLS) across all controller level objects. This is not recommended for production environments")
```
pjbgf, line 86: Let's not introduce this as breaking change now, we will revisit this in the future.
```suggestion
	flag.BoolVar(&insecureNoTLS, "insecure-no-tls", true, "Enable the use of HTTP Scheme (no TLS) across all controller level objects. This is not recommended for production environments")
```

makkes: If I understand this PR correctly than this flag's name is hard to grasp and I'd recommend this instead to clarify that this pertains to HTTP vs. HTTPS:
```suggestion
	flag.BoolVar(&insecureAllowHTTP, "insecure-allow-http", false, "Enable the use of HTTP Scheme (no HTTPS) across all controller level objects. This is not recommended for production environments")
```
Alternative something like `insecureAllowPlainHTTP` would also work.

internal/server/event_handlers.go
```diff
@@ -243,6 +244,22 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
...
 			}
 
+			webhookUrl, err := url.Parse(webhook)
+			if err != nil {
+				s.logger.Error(nil, "Error parsing webhook url",
+					"reconciler kind", v1beta1.ProviderKind,
+					"name", providerName.Name,
+					"namespace", providerName.Namespace)
+				continue
+			}
```
makkes, line 254: This case needs to be unit-tested.

internal/server/event_handlers.go
```diff
@@ -243,6 +244,22 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
...
+
+			if !s.supportHttpScheme && webhookUrl.Scheme == "http" {
+				s.logger.Error(nil, "http scheme is blocked",
+					"reconciler kind", v1beta1.ProviderKind,
+					"name", providerName.Name,
+					"namespace", providerName.Namespace)
+				continue
+			}
```
makkes, line 262: This case needs to be unit-tested.

makkes (changes requested):

gunishmatta:
@makkes  I will work on the suggested changes by you and will update the PR

Thanks for the review

stefanprodan:
@makkes I'm not Ok with breaking Flux Grafana integration by disabling http by default. If you think this is acceptable please create an RFC.

stefanprodan:
@gunishmatta please hold working on this, a decision to break Flux for all users is not something that can be taken lightly in a PR, an RFC is needed.

makkes:
> @pjbgf @makkes I'm not Ok with breaking Flux Grafana integration by disabling http by default. If you think this is acceptable please create an RFC.

My line of thought is that https://github.com/fluxcd/notification-controller/pull/404#discussion_r945833752 recommends setting the default to true for backwards-compatibility to explicitly **not** break any existing integration.

gunishmatta:
@makkes  @stefanprodan  just wanted to check if we can go ahead with making this flag disabled by default?

makkes:
> @makkes @stefanprodan just wanted to check if we can go ahead with making this flag disabled by default?

No, I don't think we can without further discussion. Disabling plain HTTP by default would break Grafana integration as @stefanprodan pointed out. An RFC would be needed.

gunishmatta:
> > @makkes @stefanprodan just wanted to check if we can go ahead with making this flag disabled by default?
> 
> No, I don't think we can without further discussion. Disabling plain HTTP by default would break Grafana integration as @stefanprodan pointed out. An RFC would be needed.

Sorry, what I mean is releasing with enabling http scheme by default.
Anyways, please let me know whatever would be the update I will be doing it accordingly.

Thanks

makkes:
> > > @makkes @stefanprodan just wanted to check if we can go ahead with making this flag disabled by default?
> > 
> > 
> > No, I don't think we can without further discussion. Disabling plain HTTP by default would break Grafana integration as @stefanprodan pointed out. An RFC would be needed.
> 
> Sorry, what I mean is releasing with enabling http scheme by default. Anyways, please let me know whatever would be the update I will be doing it accordingly.

If all comments from my and `@pjbgf`'s reviews are addressed, this should be fine to get merged.

gunishmatta:
@pjbgf  I have addressed all the changes requested above, Please review and merge if everything is okay

main.go
```diff
@@ -82,6 +83,7 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", 5*time.Minute, "Interval in which rate limit has effect.")
+	flag.BoolVar(&insecureAllowHTTP, "insecure-allow-http", false, "Enable the use of HTTP Scheme (no HTTPS) across all controller level objects. This is not recommended for production environments")
```
makkes, line 86: The default must be `true` according to [RFC 4](https://github.com/fluxcd/flux2/tree/c7e158aaa7b4c250f6c5d9fb1518a3b8bff7db36/rfcs/0004-insecure-http).

makkes (dismissed):

main.go
gunishmatta: Done

controllers/event_handling_test.go
```diff
@@ -52,7 +53,7 @@ func TestEventHandler(t *testing.T) {
 		t.Fatalf("failed to create memory storage")
 	}
 
-	eventServer := server.NewEventServer("127.0.0.1:56789", logf.Log, k8sClient, true)
+	eventServer := server.NewEventServer("127.0.0.1:56789", logf.Log, k8sClient, true, true)
```
pjbgf, line 56: Please expand the unit test to also cover when this is set to `false`.

pjbgf: @gunishmatta apart from that it is LGTM from me.

gunishmatta: @pjbgf  I have a doubt, should I create a different event server on different port ?
Something like this 	
eventServerWithHttpDisabled := server.NewEventServer("127.0.0.1:56790", logf.Log, k8sClient, true, false)
or should I create a new test method?
Please advice on this.

makkes: I'd recommend making this a table-driven test.

controllers/event_handling_test.go
```diff
@@ -52,7 +53,68 @@ func TestEventHandler(t *testing.T) {
...
+		{
+			name:          "http scheme is enabled",
+			isHttpEnabled: true,
+		}, {
+			name:          "http scheme is disabled",
+			isHttpEnabled: false,
+		},
+	}
+	for _, eventServerTest := range eventServerTests {
+
```
pjbgf, line 72:
```suggestion
```

controllers/event_handling_test.go
```diff
@@ -52,7 +53,68 @@ func TestEventHandler(t *testing.T) {
...
+				},
+				Spec: notifyv1.ProviderSpec{
+					Type:    "generic",
+					Address: rcvServer.URL,
+				},
+			}
+
+			webhook_url, err := url.Parse(provider.Spec.Address)
+			if err != nil {
+
```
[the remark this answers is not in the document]
pjbgf: At the moment this test does not assert whether the request is carried out based on the value of `supportHttpScheme`. It would be great if there was a counter inside `rcvServer` that you can use to check whether the HTTP requests get to the server when `isHttpEnabled` is `true`, but it does not when it is `false`.

controllers/event_handling_test.go
```diff
@@ -52,7 +53,68 @@ func TestEventHandler(t *testing.T) {
...
+			eventServer := server.NewEventServer("127.0.0.1:56789", logf.Log, k8sClient, true, eventServerTest.isHttpEnabled)
+
+			stopCh := make(chan struct{})
+			go eventServer.ListenAndServe(stopCh, eventMdlw, store)
+
+			rcvServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				req = r
+				w.WriteHeader(200)
+			}))
```
pjbgf, line 83: Maybe something like this:
```suggestion
			requestsReceived := 0
			rcvServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			        requestsReceived = requestsReceived + 1
				req = r
				w.WriteHeader(200)
			}))
```

And later on check for `requestsReceived` == 0 when `isHttpEnabled` is `false`. And then `requestsReceived` == 1 when `isHttpEnabled` is `true`.

controllers/event_handling_test.go
```diff
@@ -52,7 +53,69 @@ func TestEventHandler(t *testing.T) {
...
+			if err != nil {
+
+				if eventServerTest.isHttpEnabled {
+					g.Expect(webhook_url.Scheme).To(Equal(httpScheme))
+					g.Expect(requestsReceived).To(Equal(1))
+				} else {
+					g.Expect(webhook_url.Scheme).ToNot(Equal(httpScheme))
+					g.Expect(requestsReceived).To(Equal(0))
+				}
+			}
```
pjbgf, line 112: If address is not a valid URL, this would disable the checks. 
```suggestion
			webhook_url, err := url.Parse(provider.Spec.Address)
			g.Expect(err).ToNot(HaveOccurred())
			
		        if eventServerTest.isHttpEnabled {
			        g.Expect(webhook_url.Scheme).To(Equal(httpScheme))
			        g.Expect(requestsReceived).To(Equal(1))
		        } else {
			        g.Expect(webhook_url.Scheme).ToNot(Equal(httpScheme))
			        g.Expect(requestsReceived).To(Equal(0))
		        }
```

controllers/event_handling_test.go
```diff
@@ -52,7 +53,70 @@ func TestEventHandler(t *testing.T) {
...
+				},
+				Spec: notifyv1.ProviderSpec{
+					Type:    "generic",
+					Address: rcvServer.URL,
+				},
+			}
+
+			webhook_url, err := url.Parse(provider.Spec.Address)
+			g.Expect(err).ToNot(HaveOccurred())
+			if err != nil {
```
makkes, line 104: This is unnecessary given the prior call of `g.Expect`
```suggestion
```

makkes (commented):
Other than the unnecessary condition this looks good to me now.

gunishmatta:
> Other than the unnecessary condition this looks good to me now.

Done this optional change too, Please merge if everything is okay

controllers/event_handling_test.go
```diff
@@ -52,7 +53,69 @@ func TestEventHandler(t *testing.T) {
...
+			g.Expect(err).ToNot(HaveOccurred())
+
+				if eventServerTest.isHttpEnabled {
+					g.Expect(webhook_url.Scheme).To(Equal(httpScheme))
+					g.Expect(requestsReceived).To(Equal(1))
+				} else {
+					g.Expect(webhook_url.Scheme).ToNot(Equal(httpScheme))
+					g.Expect(requestsReceived).To(Equal(0))
+				}
+			}
```
pjbgf, line 112: Without the if, the code needs to be reindented:
```suggestion
		if eventServerTest.isHttpEnabled {
			g.Expect(webhook_url.Scheme).To(Equal(httpScheme))
			g.Expect(requestsReceived).To(Equal(1))
		} else {
			g.Expect(webhook_url.Scheme).ToNot(Equal(httpScheme))
			g.Expect(requestsReceived).To(Equal(0))
		}
```

pjbgf (changes requested):
@gunishmatta thank you for working on this. Once the change above is made, squash your changes to a single commit please and we should be good to merge the PR.

gunishmatta:
Hi @pjbgf , I have squashed all commits into one and it was an amazing experience contributing to Flux, Also since I am new to Golang, thanks for patiently reviewing all my changes in PR.
I would love to contribute to Flux in future too.

Thanks everyone @makkes  @stefanprodan

makkes (approved):

controllers/event_handling_test.go
```diff
@@ -173,6 +238,7 @@ func TestEventHandler(t *testing.T) {
 		res, err := http.Post("http://localhost:56789/", "application/json", buf)
 		g.Expect(err).ToNot(HaveOccurred())
 		g.Expect(res.StatusCode).To(Equal(202)) // event_server responds with 202 Accepted
+
```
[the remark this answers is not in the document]
darkowlzz: Forgot to remove the newline?

controllers/event_handling_test.go
```diff
@@ -294,4 +360,5 @@ func TestEventHandler(t *testing.T) {
 			req = nil
 		})
 	}
+
```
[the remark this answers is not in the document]
darkowlzz: Forgot to remove the newline?

controllers/event_handling_test.go
```diff
@@ -52,7 +53,69 @@ func TestEventHandler(t *testing.T) {
...
+			name:          "http scheme is enabled",
+			isHttpEnabled: true,
+		}, {
+			name:          "http scheme is disabled",
+			isHttpEnabled: false,
+		},
+	}
+	for _, eventServerTest := range eventServerTests {
+		t.Run(eventServerTest.name, func(t *testing.T) {
+
```
darkowlzz, line 73: Please initialize a new `g` here before using `g.Expect`, as it's a subtest.
Refer https://github.com/fluxcd/pkg/blob/c6f9759287b14231463a30d7ff8e416e53984a60/git/internal/e2e/gitlab_test.go#L110 for an example.

pjbgf (dismissed):

internal/server/event_server_test.go
```diff
@@ -20,29 +20,39 @@ import (
...
-	"github.com/onsi/gomega"
+	. "github.com/onsi/gomega"
 	"github.com/sethvargo/go-limiter/httplimit"
 	"github.com/sethvargo/go-limiter/memorystore"
+	"github.com/sethvargo/go-limiter/noopstore"
+	"github.com/slok/go-http-metrics/middleware"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/kubectl/pkg/scheme"
```
darkowlzz, line 37: We should be able to get rid of this new dependency by using `k8s.io/apimachinery/pkg/runtime` to create a new scheme. Refer https://github.com/fluxcd/notification-controller/blob/610fd63c11872134deb65f69c6bfaa5f00412bbd/internal/server/receiver_handler_test.go#L272

internal/server/event_handlers.go
```diff
@@ -243,6 +243,22 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
...
```
darkowlzz, line 261: Having reviewed https://github.com/fluxcd/notification-controller/pull/435, I think we can implement this block in a different way such that it becomes more apparent to the user that their Provider and Alert won't work.
At the beginning of this function, `handleEvent()`, all the alerts are listed and alerts are matched against the event. While doing so, the alerts are checked to be Ready. Not ready alerts are ignored.
In Provider reconciler, we can parse the address of the webhook and mark the object as stalled, as per the RFC https://github.com/fluxcd/flux2/tree/main/rfcs/0004-insecure-http#design-details and Ready=False. But if the address is specified in a secret ref, the address in the secret can change without updating the Provider object. So, when secretRef is present, we can just fail the reconciliation with Ready=False and allow it to retry with exponential backoff.
Because the Provider is not ready, the associated Alert would also become not ready and that intern would make the event handler to drop the event early in the above function. The failure in the configuration would be visible on the object itself, compared to the current implementation where it'll be just logged and may not be visible.

stefanprodan: I propose we delay this PR until v1beta2 is released. If RFC-0004 instructs the objects to marked as stalled, then we'll probably need to add secret watches to NC and cascade stalling from Provider to all dependant Alerts.

pjbgf: @gunishmatta unfortunately, we will have to label these changes on hold until the #435 is merged, by which point we will need to review the implementation based on @darkowlzz comments above.

gunishmatta: No Worries, Will keep following it and would focus on contributing to other issues at Flux and understanding the code in depth.

Issue #418, opened 2022-09-12
Request - exclusionList in Alert should filter on more than just the message


philipsabri:
Hello! Currently running into a few problems with more "complex" `Alert` configs.

Would be nice if we could use exclusionList on more than just the message such as the name of the alert, which I believe is called `event.InvolvedObject.Name`

I would like to run something like this without having to hardcore all services in `eventSources` for the first `Alert` config but still be able to add more customizable slack messages on only some services.

```
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: service-deploy
  namespace: flux-system
spec:
  summary: "HelmRelease deploys"
  providerRef:
    name: slack-service-deploy
  eventSeverity: info
  eventSources:
    - kind: HelmRelease
      name: "*"
  exclusionList:
    - ".*flurrens-service.*"
---
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: service-deploy-flurren
  namespace: flux-system
spec:
  summary: "tex text notify owner <@SLACK_ID_HERE>"
  providerRef:
    name: slack-service-deploy
  eventSeverity: info
  eventSources:
    - kind: HelmRelease
      name: "*"
      matchLabels:
        owner: flurren
```

flurreN:
Made a PR if this could be something https://github.com/fluxcd/notification-controller/pull/419 I dont know GO so it could probably be done in other ways.

PR #419, opened 2022-09-12
Be able to exclude event names in Alert exclusionList

philipsabri:
This adds the possibility to exclude alerts based on event name

Fixes https://github.com/fluxcd/notification-controller/issues/418


stefanprodan:
If we are making breaking changes I would suggest correcting the variables names, some of them are not capitalized correctly for example `serviceaccount` should be `serviceAccount`, `additionalargs` should be `additionalArgs`, `watchallnamespaces` should be `watchAllNamespaces`.

PR #423, opened 2022-09-26
Update dependencies

stefanprodan:
Updates:
- k8s.io/* v0.25.2
- controller-runtime v0.13.0
- controller-gen v0.8.0
- fluxcd/pkg/runtime v0.19.0
- Go 1.19


controllers/receiver_controller.go
```diff
@@ -159,7 +159,7 @@ func (r *ReceiverReconciler) SetupWithManagerAndOptions(mgr ctrl.Manager, opts R
 		Complete(r)
 }
 
-/// reconcile steps through the actual reconciliation tasks for the object, it returns early on the first step that
+// / reconcile steps through the actual reconciliation tasks for the object, it returns early on the first step that
```
hiddeco, line 162:
```suggestion
// reconcile steps through the actual reconciliation tasks for the object, it returns early on the first step that
```

hiddeco (approved):
Other than the commented nit, this looks good to me.

PR #424, opened 2022-09-27
Dockerfile: Build with Go 1.19

stefanprodan:
Update build image to Go 1.19


hiddeco (commented):
Same comment as https://github.com/fluxcd/image-reflector-controller/pull/317#discussion_r981243301

PR #426, opened 2022-09-30
Add "generic-hmac" Provider

### LINKED ISSUE #99: Allow signing of generic webhook provider requests

makkes:
This commit adds the "generic-hmac" Provider type for authenticating
webhook requests coming from notification-controller. I extended the
`Forwarder` notifier to accept an optional key used for generating the
HMAC. If the key is nil or empty no HMAC header is generated and the
forwarder behaves as before. If it is provided an `X-Signature` HTTP
header is added to the request carrying the HMAC.

I transformed the `TestForwarder_Post` test into a table-driven test
so that we can use the same setup and testing code for testing HMAC
and non-HMAC forwarder instances.

The Flux CLI will support this new type automatically as it doesn't validate the given type by itself (this is done by the API server using the CRD spec.

closes #99

### DIFF
--- a/internal/notifier/forwarder.go
+++ b/internal/notifier/forwarder.go
@@ -18,7 +18,10 @@ package notifier
 
 import (
 	"context"
+	"crypto/hmac"
+	"crypto/sha256"
 	"crypto/x509"
+	"encoding/json"
 	"fmt"
 	"net/url"
 
@@ -38,11 +41,16 @@ type Forwarder struct {
 	ProxyURL string
 	Headers  map[string]string
 	CertPool *x509.CertPool
+	HMACKey  []byte
 }
 
-func NewForwarder(hookURL string, proxyURL string, headers map[string]string, certPool *x509.CertPool) (*Forwarder, error) {
+func NewForwarder(hookURL string, proxyURL string, headers map[string]string, certPool *x509.CertPool, hmacKey []byte) (*Forwarder, error) {
 	if _, err := url.ParseRequestURI(hookURL); err != nil {
 		return nil, fmt.Errorf("invalid hook URL %s: %w", hookURL, err)
+	}
+
+	if hmacKey != nil && len(hmacKey) == 0 {
+		return nil, fmt.Errorf("HMAC key is empty")
 	}
 
 	return &Forwarder{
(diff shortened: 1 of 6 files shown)


makkes:
Sorry, I broke the unit tests with my latest change. Will fix them right away.

internal/notifier/forwarder.go
```diff
@@ -38,23 +41,46 @@ type Forwarder struct {
...
+		eventJSON, err := json.Marshal(event)
+		if err != nil {
+			return fmt.Errorf("failed marshalling event: %w", err)
+		}
+		sig = fmt.Sprintf("sha256=%s", sign(eventJSON, f.HMACKey))
+	}
 	err := postMessage(ctx, f.URL, f.ProxyURL, f.CertPool, event, func(req *retryablehttp.Request) {
+		if sig != "" {
+			req.Header.Set("X-Signature", sig)
+		}
```
pjbgf, line 83: We should move this to after the `for` loop, to ensure that the signature created by setting the `f.HMACKey` would not be overwritten by `f.Headers` having a header with the same name. Or have validation to error in such situation.

internal/notifier/forwarder_test.go
```diff
@@ -25,42 +25,113 @@ import (
...
+		{
+			name:       "happy path with nil HMAC key",
+			hmacKey:    nil,
+			hmacHeader: "",
+		},
+		{
+			name:       "non-empty HMAC key adds signature header",
+			hmacKey:    []byte("7152fed34dd6149a7c75a276c510da27cb6f82b0"),
+			hmacHeader: "sha256=65b018549b1254e7226d1c08f9567ee45bc9de0fc4e7b1a40253f9a018b08be7",
+		},
```
pjbgf, line 92: We could add a test to ensure that the headers input does not overwrite the signature created by the new code.

makkes: Great idea. I added some more testing around the `X-Signature` header.

pjbgf (approved):
LGTM

Thanks @makkes! :bow:

PR #427, opened 2022-10-06
Prepend spec.summary to the github status message

alexkashintsev-tomtom:
Prepend Alert: spec.summary to the name of github notification.

I would like to use Flux notification to github which will show the badge on successfull/unsuccessfull reconciliation.
But I would like to modify message and include name of the region which was able to apply changes.

Currently Flux notification controller does not take into account configured summary of the alert.
By default if there is no summary in the metadata it will generate default status message.

if summary is provided Notifications controller will prepend it to the name of status: by doing this multiple regions can report to the same status in Github.


alexkashintsev-tomtom:
Feel like dup of #340

PR #429, opened 2022-10-15
Add Terraform event source kind

k0da:
Signed-off-by: Dinar Valeev <dinar.valeev@absa.africa>

### DIFF
--- a/api/v1beta1/reference_types.go
+++ b/api/v1beta1/reference_types.go
@@ -24,7 +24,7 @@ type CrossNamespaceObjectReference struct {
 	APIVersion string `json:"apiVersion,omitempty"`
 
 	// Kind of the referent
-	// +kubebuilder:validation:Enum=Bucket;GitRepository;Kustomization;HelmRelease;HelmChart;HelmRepository;ImageRepository;ImagePolicy;ImageUpdateAutomation;OCIRepository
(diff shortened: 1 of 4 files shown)


stefanprodan:
The Terraform Kind is not an a Flux CRD, if you want  to enable notifications for 3rd party controllers see https://fluxcd.io/flux/cheatsheets/bootstrap/#enable-notifications-for-third-party-controllers

k0da:
Sorry, I don't have kustomize controller deployed.  Wouldn't it be easier to relax validation here? 
It is a little bit an overkill to add a new type.

stefanprodan:
@k0da you do a `kubectl patch` for each patch in https://fluxcd.io/flux/cheatsheets/bootstrap/#enable-notifications-for-third-party-controllers, his is not only about adding Terraform to the CRD but also about RBAC.

stefanprodan:
> Wouldn't it be easier to relax validation here?

The validation is only half the problem, notification-controller needs to have read access to that Kind at cluster level too. We have no plans on given this controller cluster-admin, or allow it to read any random object from the cluster.

k0da:
@stefanprodan but this can be fixed on tf-controller side. A chart can bring a role and binding.

PR #442, opened 2022-11-22
Update dependencies 

hiddeco:


pjbgf:
We can ignore the fuzz errors as it broke due to changes upstream, I will pick that up separately.

hiddeco:
@stefanprodan please take a good look at the latest commit, as it includes some minor changes from your thing that touch backwards compatibility related things. If these should wait until next API version, I can add them back.

stefanprodan (approved):
LGTM

PR #448, opened 2022-12-01
receiver: Use new ServeMux

darkowlzz:
Receiver should use its own ServeMux.

Related to https://github.com/fluxcd/notification-controller/issues/449


pjbgf (approved):
LGTM

Thanks @darkowlzz. :bow:

somtochiama (approved):
Thanks @darkowlzz !

PR #451, opened 2022-12-04
feat: add gitea notifier

ttys3:
add support to send event notification to Gitea repo commit status

image

Gitea is a forge software package for hosting software development version control using Git as well as other collaborative features like bug tracking, code review, kanban boards, tickets, and wikis. It supports self-hosting but also provides a free public first-party instance

https://github.com/go-gitea/gitea

### DIFF
--- a/internal/notifier/gitea.go
+++ b/internal/notifier/gitea.go
@@ -1,0 +1,192 @@
+/*
+Copyright 2022 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package notifier
+
+import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
+	"errors"
+	"fmt"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+
+	"code.gitea.io/sdk/gitea"
+	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/fluxcd/pkg/apis/meta"
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+type Gitea struct {
+	BaseURL     string
+	Token       string
+	Owner       string
+	Repo        string
+	ProviderUID string
+	Client      *gitea.Client
+	Debug       bool
+}
+
+var _ Interface = &Gitea{}
+
+func NewGitea(providerUID string, addr string, token string, certPool *x509.CertPool) (*Gitea, error) {
+	if len(token) == 0 {
+		return nil, errors.New("gitea token cannot be empty")
+	}
+
+	host, id, err := parseGitAddress(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := url.Parse(host); err != nil {
+		return nil, err
+	}
+
+	idComponents := strings.Split(id, "/")
+	if len(idComponents) != 2 {
+		return nil, fmt.Errorf("invalid repository id %q", id)
+	}
+
+	client, err := gitea.NewClient(host, gitea.SetToken(token))
+	if err != nil {
+		return nil, err
+	}
+
+	if certPool != nil {
+		tr := &http.Transport{
+			TLSClientConfig: &tls.Config{
+				RootCAs: certPool,
+			},
+		}
+		client.SetHTTPClient(&http.Client{Transport: tr})
+	}
+
+	return &Gitea{
+		BaseURL:     host,
+		Token:       token,
+		Owner:       idComponents[0],
+		Repo:        idComponents[1],
+		ProviderUID: providerUID,
+		Client:      client,
+		Debug:       os.Getenv("NOTIFIER_GITEA_DEBUG") == "true",
+	}, err
+}
+
+func (g *Gitea) Post(ctx context.Context, event eventv1.Event) error {
+	revString, ok := event.Metadata[eventv1.MetaRevisionKey]
+	if !ok {
+		return errors.New("missing revision metadata")
+	}
+	rev, err := parseRevision(revString)
+	if err != nil {
+		return err
+	}
+	state, err := toGiteaState(event)
+	if err != nil {
+		return err
+	}
+
+	_, desc := formatNameAndDescription(event)
+	id := generateCommitStatusID(g.ProviderUID, event)
+
+	status := gitea.CreateStatusOption{
+		State:       state,
+		TargetURL:   "",
+		Description: desc,
+		Context:     id,
+	}
+
+	listStatusesOpts := gitea.ListStatusesOption{
+		ListOptions: gitea.ListOptions{
+			Page:     0,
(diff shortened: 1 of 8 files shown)


internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
+package notifier
+
+import (
```
makkes, line 3: Please arrange the imports here the same way as in the other files of this package. https://github.com/golang/go/wiki/CodeReviewComments#imports

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+	"strings"
+)
+
+type Gitea struct {
+	BaseURL string
+	Token   string
+	Owner   string
+	Repo    string
+	client  *gitea.Client
+	debug   bool
```
makkes, line 25: These should be public as well.

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+var _ Interface = &Gitea{}
+
+func NewGitea(addr string, token string, certPool *x509.CertPool) (*Gitea, error) {
```
makkes, line 30:
```suggestion
func NewGitea(addr, token string, certPool *x509.CertPool) (*Gitea, error) {
```

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+	if len(token) == 0 {
+		return nil, errors.New("github token cannot be empty")
```
makkes, line 32:
```suggestion
		return nil, errors.New("gitea token cannot be empty")
```

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+		debug = true
+	}
+
+	return &Gitea{
+		BaseURL: host,
+		Token:   token,
+		Owner:   comp[0],
+		Repo:    comp[1],
+		client:  client,
+		debug:   debug,
```
makkes, line 74:
```suggestion
	return &Gitea{
		BaseURL: host,
		Token:   token,
		Owner:   comp[0],
		Repo:    comp[1],
		client:  client,
		debug:   os.Getenv("NOTIFIER_GITEA_DEBUG") == "true",
```

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+	host, id, err := parseGitAddress(addr)
+	if err != nil {
+		return nil, err
+	}
+
+	if _, err := url.Parse(host); err != nil {
+		return nil, err
+	}
+
+	comp := strings.Split(id, "/")
```
makkes, line 44:
```suggestion
	idComponents := strings.Split(id, "/")
```

internal/notifier/gitea.go
```diff
@@ -0,0 +1,176 @@
...
+
+	listStatusesOpts := gitea.ListStatusesOption{
+		ListOptions: gitea.ListOptions{
+			Page:     0,
+			PageSize: 50,
+		},
+	}
+	statuses, _, err := g.client.ListStatuses(g.Owner, g.Repo, rev, listStatusesOpts)
+	if err != nil {
+		return fmt.Errorf("could not list commit statuses: %v", err)
```
makkes, line 108:
```suggestion
		return fmt.Errorf("could not list commit statuses: %w", err)
```

internal/notifier/gitea_test.go
```diff
@@ -0,0 +1,41 @@
+/*
+Copyright 2020 The Flux authors
```
makkes, line 2:
```suggestion
Copyright 2022 The Flux authors
```

internal/notifier/gitea_test.go
```diff
@@ -0,0 +1,41 @@
+/*
```
makkes, line 1: Please also test the `Post` method appropriately.

makkes (changes requested):

ttys3:
the tests failed.  it complains about `config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml` which I added `gitea`

kingdonb:
I have an idea @ttys3 – it looks like you need to run `make generate` and commit the result. The error is a "dirty working tree" which just means that something in the e2e job runs `make generate` and you're supposed to already have run this.

I tried that locally but it only came up with half of the changes that failing test mentioned, the other one in `go.sum`

I guess that `make test` runs all of that

I found both the change in `go.sum` and in `notification[..]providers.yaml` like you mentioned. If you run the tests locally, you should get those changes in your working tree then you can commit them and e2e should hopefully pass after that 🎉

ttys3:
@kingdonb `make test` will remove `gitea` from config/crd/bases/notification.toolkit.fluxcd.io_providers.yaml

`make generate` is right.

I need add `gitea` to api/v1beta1/provider_types.go, the yaml is generated from the annotation

ttys3:
OK. fixed

PTAL

makkes (approved):

makkes:
@ttys3 please squash the commits into one.

ttys3:
squash done.

makkes (approved):

makkes (dismissed):
@stefanprodan actually reminded me that we need to update the docs accordingly here. @ttys3 please add documentation about this new type to [the provider documentation page](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta1/provider.md).

ttys3:
@makkes added. PTAL

stefanprodan:
@ttys3 this need to be ported to the v1beta2 API, we've deprecated v1beta1 in the main branch and that version is now freezed.

api/v1beta1/provider_types.go
```diff
@@ -30,7 +30,7 @@ const (
 // ProviderSpec defines the desired state of Provider
 type ProviderSpec struct {
 	// Type of provider
-	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;generic-hmac;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;opsgenie;alertmanager;grafana;githubdispatch;
+	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;generic-hmac;github;gitlab;gitea;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;opsgenie;alertmanager;grafana;githubdispatch;
```
stefanprodan, line 33: Please undo all the changes to v1beta1 and move them to v1beta2, then run codegen for the API docs.

docs/spec/v1beta1/provider.md
```diff
@@ -9,7 +9,7 @@ Spec:
 ```go
 type ProviderSpec struct {
 	// Type of provider
-	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;generic-hmac;github;gitlab;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;opsgenie;githubdispatch
+	// +kubebuilder:validation:Enum=slack;discord;msteams;rocket;generic;generic-hmac;github;gitlab;gitea;bitbucket;azuredevops;googlechat;webex;sentry;azureeventhub;telegram;lark;matrix;opsgenie;githubdispatch
```
stefanprodan, line 12: Please undo this change and add Gitea to the v1beta2 docs including a section about auth, like this https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta2/providers.md#github

ttys3:
CI failed. force pushed.
PTAL

stefanprodan (approved):
LGTM 

Thanks @ttys3 🥇

makkes (approved):
Great job 🎖️

PR #452, opened 2022-12-13
controllers: Get object before Update in tests

hiddeco:


stefanprodan (approved):
LGTM

Thanks @hiddeco

PR #455, opened 2022-12-21
Fix panic when upgrading to v1beta2

stefanprodan:
The Kubernetes conversion webhook doesn't set the default value for the newly added `interval` field, to avoid controller panic, we set the default in code.

PR #457, opened 2022-12-22
Remove interval default value from CRDs

stefanprodan:
Fix: https://github.com/fluxcd/flux2/discussions/3434
Followup: https://github.com/fluxcd/notification-controller/pull/455


relu (approved):
LGTM

PR #482, opened 2023-03-06
feat: support multi receiver by matchLabels

raffis:
## Current situation

At the moment one has to reference each resource by name for a webhook receiver.
While spec.resources actually is a `CrossNamespaceObjectReference` which supports matchLabels. However this is not implemented in the internal handling of the receiver.  What I like to achieve is basically trigger a reconcile for multiple HelmCharts matching a certain label (or just all of them in a specific namespace). 

This currently is unsupported in the reconcile logic while it is supported in the api:
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta2
kind: Receiver
metadata:
  name: ghcr-receiver
  namespace: flux-system
spec:
  events:
  - package
  resources:
  - apiVersion: source.toolkit.fluxcd.io/v1beta2
    kind: HelmChart
    matchLabels:
      registry: ghcr
    name: '*'
  secretRef:
    name: github-webhook-token
  type: github
```

## Background

I like to send a webhook from github for helm packages published in ghcr. The receiver should reconcile all (or matching helm charts by label) charts and trigger a reconcile.
What I am trying to avoid is set a very low interval but instead reconcile a helmrelease immediately if a new version is published via webhook.
This works perfectly fine already but I would need to specify each chart manually in the list of resources.

## Proposal
Support matchLabels and `*` as name the same way as this is already supported for Alerts in the notification-controller.

### DIFF
--- a/internal/server/receiver_handler_test.go
+++ b/internal/server/receiver_handler_test.go
@@ -563,28 +563,7 @@ func Test_handlePayload(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			resources: []client.Object{
-				&apiv1.Receiver{
-					TypeMeta: metav1.TypeMeta{
-						Kind:       apiv1.ReceiverKind,
-						APIVersion: apiv1.GroupVersion.String(),
-					},
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "dummy-resource-2",
-					},
-				},
-				&apiv1.Receiver{
-					TypeMeta: metav1.TypeMeta{
-						Kind:       apiv1.ReceiverKind,
-						APIVersion: apiv1.GroupVersion.String(),
-					},
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "dummy-resource",
-					},
-				},
-			},
-			expectedResourcesAnnotated: 3, // it is 3 because we target receivers itself in this test and the receiver here is included with *
-			expectedResponseCode:       http.StatusOK,
+			expectedResponseCode: http.StatusBadRequest,
 		},
 		{
 			name: "resource matchLabels is ignored if name is not *",
--- a/internal/server/receiver_handler_test.go
+++ b/internal/server/receiver_handler_test.go
@@ -23,13 +23,16 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"net/http"
 	"net/http/httptest"
 	"testing"
 
 	"github.com/google/go-github/v41/github"
+	"github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/fluxcd/pkg/apis/meta"
@@ -38,21 +41,23 @@ import (
 	apiv1 "github.com/fluxcd/notification-controller/api/v1beta2"
 )
 
-func Test_validate(t *testing.T) {
+func Test_handlePayload(t *testing.T) {
 	type hashOpts struct {
 		calculate bool
 		header    string
 	}
 
 	tests := []struct {
-		name         string
-		hashOpts     hashOpts
-		headers      map[string]string
-		payload      map[string]interface{}
-		receiver     *apiv1.Receiver
-		receiverType string
-		secret       *corev1.Secret
-		expectedErr  bool
+		name                       string
+		hashOpts                   hashOpts
+		headers                    map[string]string
+		payload                    map[string]interface{}
+		receiver                   *apiv1.Receiver
+		receiverType               string
+		secret                     *corev1.Secret
+		resources                  []client.Object
+		expectedResourcesAnnotated int
+		expectedResponseCode       int
 	}{
 		{
 			name: "Generic receiver",
@@ -66,6 +71,10 @@ func Test_validate(t *testing.T) {
 						Name: "token",
 					},
 				},
+				Status: apiv1.ReceiverStatus{
+					WebhookPath: apiv1.ReceiverWebhookPath,
+					Conditions:  []metav1.Condition{{Type: meta.ReadyCondition, Status: metav1.ConditionTrue}},
+				},
 			},
 			secret: &corev1.Secret{
 				ObjectMeta: metav1.ObjectMeta{
@@ -75,7 +84,7 @@ func Test_validate(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedErr: false,
+			expectedResponseCode: http.StatusOK,
 		},
 		{
 			name: "gitlab receiver",
@@ -89,6 +98,10 @@ func Test_validate(t *testing.T) {
 						Name: "token",
 					},
 				},
+				Status: apiv1.ReceiverStatus{
+					WebhookPath: apiv1.ReceiverWebhookPath,
+					Conditions:  []metav1.Condition{{Type: meta.ReadyCondition, Status: metav1.ConditionTrue}},
+				},
 			},
 			headers: map[string]string{
 				"X-Gitlab-Token": "token",
@@ -101,7 +114,7 @@ func Test_validate(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedErr: false,
+			expectedResponseCode: http.StatusOK,
 		},
 		{
 			name: "github receiver",
@@ -114,6 +127,10 @@ func Test_validate(t *testing.T) {
 					SecretRef: meta.LocalObjectReference{
 						Name: "token",
 					},
+				},
+				Status: apiv1.ReceiverStatus{
+					WebhookPath: apiv1.ReceiverWebhookPath,
+					Conditions:  []metav1.Condition{{Type: meta.ReadyCondition, Status: metav1.ConditionTrue}},
 				},
 			},
 			hashOpts: hashOpts{
@@ -134,7 +151,7 @@ func Test_validate(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedErr: false,
+			expectedResponseCode: http.StatusOK,
 		},
 		{
 			name: "generic hmac receiver",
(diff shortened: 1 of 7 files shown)


internal/server/receiver_handlers.go
```diff
@@ -378,15 +404,25 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource apiv1.CrossNames
 		return fmt.Errorf("unable to read %s '%s' error: %w", resource.Kind, objectKey, err)
 	}
```
somtochiama, line 406: Need to call `s.annotate` for the single object at the end.

It would be great to add some tests for the function using the fake client.

internal/server/receiver_handlers.go
```diff
@@ -367,6 +369,30 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource apiv1.CrossNames
 
 	group, version := getGroupVersion(apiVersion)
 
+	if resource.Name == "*" && resource.MatchLabels != nil {
```
somtochiama, line 372:
```suggestion
	if resource.Name == “” && resource.MatchLabels != nil {
```

I don’t think we need to support wildcard in the name but the name should take precedence over matchLabels if both are set

>

somtochiama (changes requested):

internal/server/receiver_handlers.go
raffis: I agree with that, it's done like that to keep things consistent how things are done for `Alert`  https://github.com/fluxcd/notification-controller/blob/main/internal/server/event_handlers.go#L293

internal/server/receiver_handlers.go
raffis: ouch, good 🎣

makkes (changes requested):
I don't understand how this is supposed to work. The `.name` field is still mandatory and this PR doesn't change the way it is handled. The PR description says you can set `name` to `*` the same way we support it with Alerts but the commits don't back this up.

In fact, `name` is still used to fetch a single resource and in addition `matchLabels` is taken into account as well.

raffis:
> I don't understand how this is supposed to work. The `.name` field is still mandatory and this PR doesn't change the way it is handled. The PR description says you can set `name` to `*` the same way we support it with Alerts but the commits don't back this up.
> 
> In fact, `name` is still used to fetch a single resource and in addition `matchLabels` is taken into account as well.

Hey please align on your side and decide what you want, https://github.com/fluxcd/notification-controller/pull/482#discussion_r1132222561

makkes:
> > I don't understand how this is supposed to work. The `.name` field is still mandatory and this PR doesn't change the way it is handled. The PR description says you can set `name` to `*` the same way we support it with Alerts but the commits don't back this up.
> > In fact, `name` is still used to fetch a single resource and in addition `matchLabels` is taken into account as well.
> 
> Hey please align on your side and decide what you want, [#482 (comment)](https://github.com/fluxcd/notification-controller/pull/482#discussion_r1132222561)

I see 2 issues with the details of this PR:

1. The `.spec.resources.name` field is required so it can never be empty.
2. Even if we changed it back to `*` as a special case (sorry, haven't looked at the resolved comments) the receiver handler would still try to get an object of that name and fail.

A solution would be to accept `*` as a name and treat it specially. However, there's still two cases that would need to be covered:

- `.spec.resources.name` == `*` and `.spec.resources.matchLabels` is missing.
- `.spec.resources.name` == `*` and `.spec.resources.matchLabels` is present.

raffis:
> 



> > > I don't understand how this is supposed to work. The `.name` field is still mandatory and this PR doesn't change the way it is handled. The PR description says you can set `name` to `*` the same way we support it with Alerts but the commits don't back this up.
> > > In fact, `name` is still used to fetch a single resource and in addition `matchLabels` is taken into account as well.
> > 
> > 
> > Hey please align on your side and decide what you want, [#482 (comment)](https://github.com/fluxcd/notification-controller/pull/482#discussion_r1132222561)
> 
> I see 2 issues with the details of this PR:
> 
>     1. The `.spec.resources.name` field is required so it can never be empty.
> 
>     2. Even if we changed it back to `*` as a special case (sorry, haven't looked at the resolved comments) the receiver handler would still try to get an object of that name and fail.

I will change it back then. That's what I suggested in the first place.

> 
> A solution would be to accept `*` as a name and treat it specially. However, there's still two cases that would need to be covered:
> 
>     * `.spec.resources.name` == `*` and `.spec.resources.matchLabels` is missing.

Either we can can change it so if no matchLabels are set and its `*` it would select all resources. Or if no change it would just fallback to select by name which would fail to get a resource since there is none named `*`. I would say first one.


>     * `.spec.resources.name` == `*` and `.spec.resources.matchLabels` is present.

Well thats the case which is covered if I change it back how it was initially.



> But then the question arises of how we handle `.spec.resources.name` not being `*` and `.spec.resources.matchLabels` being present. This case doesn't make sense and IMHO should be prevented at the validation level.

This would require a validation hook I reckon. But we could also argue in the current version you can also set * and matchLabels to anything and nothing happens since it's not supported.

somtochiama:
@makkes My bad, I suggested checking for an empty name field in my last review, forgetting that the name was required.
Thanks @raffis for reverting the change

internal/server/receiver_handlers.go
```diff
@@ -94,7 +95,7 @@ func (s *ReceiverServer) handlePayload() func(w http.ResponseWriter, r *http.Req
 			}
 
 			for _, resource := range receiver.Spec.Resources {
-				if err := s.annotate(ctx, resource, receiver.Namespace); err != nil {
+				if err := s.annotateCrossNamespaceReferences(ctx, resource, receiver.Namespace); err != nil {
```
somtochiama, line 98: Testing this pr, in the logs of the notification controller I see
```
{"level":"info","ts":"2023-03-13T19:54:31.675+0100","logger":"receiver-server","msg":"resource 'ImageRepository/*.default' annotated","reconciler kind":"Receiver","name":"webapp","namespace":"flux-system"}
```

which isn't super informative (It gives this log even when no resources with the labels were found).
I think we might need to pass the logger down to the `annotateCrossNamespaceReferences` to give better logs when each resource is annotated (Wdyt? cc @makkes )

raffis: Agreed, redone logging a bit, pls check again 🙏🏻

somtochiama: Any reason why you are loggin at debug level? The log doesn't show up with the default log option
```
s.logger.V(1).Info(fmt.Sprintf("resource '%s/%s.%s' annotated",
					resource.Kind, resource.Name, namespace))
```

Secondly,  the log no longer has the receiver(name, ns) details associated with it. I am thinking we might want to pass in the logger created with these details into the function rather than use the one in the struct. It would help to differentiate logs for one receiver from another.

raffis: Yes on purpose because this could target many resources and could be logged multiple times. There is a new log before individual resources. Let me know if it should be changed to info.

raffis: >  I am thinking we might want to pass in the logger created with these details into the function rather than use the one in the struct. It would help to differentiate logs for one receiver from another.

yes makes sense, it's changed now

internal/server/receiver_handler_test.go
```diff
@@ -265,8 +298,234 @@ func Test_validate(t *testing.T) {
...
+					Name: "token",
+				},
+				Data: map[string][]byte{
+					"token": []byte("token"),
+				},
+			},
+			expectedResponseCode: http.StatusBadRequest,
+		},
+		{
+			name: "annotating resources by label match",
```
makkes, line 412: please add tests for the edge cases:

- Name == '*' and MatchLables is nil/empty
- Name != '*' and != '' and MatchLabels is present

internal/server/receiver_handlers.go
```diff
@@ -347,15 +345,11 @@ func (s *ReceiverServer) token(ctx context.Context, receiver apiv1.Receiver) (st
 	return token, nil
 }
 
-func (s *ReceiverServer) annotate(ctx context.Context, resource apiv1.CrossNamespaceObjectReference, defaultNamespace string) error {
+func (s *ReceiverServer) annotateCrossNamespaceReferences(ctx context.Context, logger logr.Logger, resource apiv1.CrossNamespaceObjectReference, defaultNamespace string) error {
```
makkes, line 348:
```suggestion
// requestReconciliation requests reconciliation of all the resources matching the given CrossNamespaceObjectReference by annotating them accordingly.
func (s *ReceiverServer) requestReconciliation(ctx context.Context, logger logr.Logger, resource apiv1.CrossNamespaceObjectReference, defaultNamespace string) error {
```

makkes (commented):
This looks much better now. Just some suggestions and please also update docs/api/notification.md to reflect the changes around the semantics of `.spec.resources`.

internal/server/receiver_handler_test.go
raffis: > Name == '*' and MatchLables is nil/empty
@makkes there is still that open question if name is `*` and not matchLabels are defined if it should target all resources matching the kind? I would opt for yes.

> Name != '*' and != '' and MatchLabels is present
In that case we agree matchLabel would just be ignored right? And It would just get the resource by name.

makkes: For Alerts if Name == '*' then it matches any source so I agree we should implement the same behaviour here as well.

Also for Alerts if Name != '*' and MatchLabels != nil then, yes, MatchLabels is just ignored and we should replicate that behaviour.

internal/server/receiver_handlers.go
```diff
@@ -367,26 +362,79 @@ func (s *ReceiverServer) annotate(ctx context.Context, resource apiv1.CrossNames
...
+			Group:   group,
+			Kind:    resource.Kind,
+			Version: version,
+		})
+
+		if err := s.kubeClient.List(ctx, &resources,
+			client.InNamespace(namespace),
+			client.MatchingLabels(resource.MatchLabels),
+		); err != nil {
+			return err
```
makkes, line 380: As we're not logging the resource details in the caller, anymore, we need to add these here:
```suggestion
			return fmt.Errorf("failed listing resources in namespace %q by matching labels %q: %w", namespace, err)
```

makkes (approved):
I think we have everything in place right now. Good job, @raffis. 🎖️ 

Please squash all your commits into one, then I'll make sure to get it merged.

raffis:
> I think we have everything in place right now. Good job, @raffis. medal_military
> 
> Please squash all your commits into one, then I'll make sure to get it merged.

Done, thx for the review 🥇

makkes:
Holding for next minor release.

makkes (changes requested):
I'm sorry @raffis but I need to ask for one additional change: Please document the changed API in https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta2/receivers.md#resources. The content of that page is published at https://fluxcd.io/flux/components/notification/receiver/ so we need give clear guidance on how to use `matchLabels`.

makkes (approved):

PR #504, opened 2023-03-30
Allow wildcard value for namespaces in Alert eventSources

### LINKED ISSUE #71: Add possibility to configure wildcard namespace value in Alert eventSources
It would be nice to pick up all error events within a cluster using a wildcard namespace value in the eventSources section of an Alert  - similar to how it works for name:
https://github.com/fluxcd/notification-controller/blob/fbf1ea0413e12fe58e6386972468a152c42b215c/internal/server/event_handlers.go#L77

Currently you'll either need to duplicate Provider (including any secrets) and Alert resources to all relevant namespaces or create a "global" Alert which references other namespaces. It would be simpler and less error prone to allow `namespace: "*"`.

alekspog:
Allow wildcard value for namespaces in Alert eventSources.

Fix: https://github.com/fluxcd/notification-controller/issues/71

At the moment we have to create a list of all HelmRelease objects, for example, in the `eventSources` section to be able to get notifications about them. We also need to update the list whenever a new object is added. That requires lots of effort. It seems that the best way is to allow wildcards for namespaces.

How the Alert `eventSources` definition looks now:

```yaml
  eventSources:
    - kind: HelmRelease
      name: '*'
      namespace: 'service1'
    - kind: HelmRelease
      name: '*'
      namespace: 'service2'
...
    - kind: HelmRelease
      name: '*'
      namespace: 'service99'
```

How it would be after the change:

```yaml
  eventSources:
    - kind: HelmRelease
      name: '*'
      namespace: '*'
```

### DIFF
--- a/docs/spec/v1beta2/alerts.md
+++ b/docs/spec/v1beta2/alerts.md
@@ -142,6 +142,19 @@ eventSources:
     namespace: apps
 ```
 
+#### Select objects from all namespaces
+
+The `*` wildcard can be used to select events issued by Flux objects from any `namespace`:
+
+```yaml
+eventSources:
+  - kind: HelmRelease
+    name: 'service1'
+    namespace: '*'
+```
+
+It requires to have cross-namespace references enabled for the controller.
+
 #### Select objects by label
 
 To select events issued by all Flux objects of a particular `kind` with specific `labels`:
--- a/go.mod
+++ b/go.mod
@@ -24,6 +24,7 @@ require (
 	github.com/ktrysmt/go-bitbucket v0.9.55
 	github.com/microsoft/azure-devops-go-api/azuredevops/v6 v6.0.1
 	github.com/onsi/gomega v1.27.2
+	github.com/prometheus/client_golang v1.14.0
 	github.com/sethvargo/go-limiter v0.7.2
 	github.com/slok/go-http-metrics v0.10.0
 	github.com/spf13/pflag v1.0.5
@@ -110,7 +111,6 @@ require (
 	github.com/peterbourgon/diskv v2.0.1+incompatible // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
-	github.com/prometheus/client_golang v1.14.0 // indirect
 	github.com/prometheus/client_model v0.3.0 // indirect
 	github.com/prometheus/common v0.37.0 // indirect
 	github.com/prometheus/procfs v0.8.0 // indirect
--- a/internal/controllers/alert_controller_test.go
+++ b/internal/controllers/alert_controller_test.go
@@ -267,11 +267,6 @@ func TestAlertReconciler_EventHandler(t *testing.T) {
 					Name:      "*",
 					Namespace: "test",
 				},
-				{
-					Kind:      "Kustomization",
-					Name:      "testwildcardnamespace",
-					Namespace: "*",
-				},
 			},
 			ExclusionList: []string{
 				"doesnotoccur", // not intended to match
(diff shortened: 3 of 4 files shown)


makkes:
There are security concerns related to this feature that were articulated in #71 (e.g. https://github.com/fluxcd/notification-controller/issues/71#issuecomment-991049078). Given the mentioned RFC has never been merged, we need to think this over.

In any case this will need to be held until notification-controller 1.0.0 has been released. Let's please take the time to completely discuss the threat model and potential mitigations here.

alekspog:
I thought that disabling cross-namespace references should solve the security concern as was said there:
https://github.com/fluxcd/notification-controller/issues/71#issuecomment-1374690404. I agree that using [ReferenceGrant](https://github.com/fluxcd/flux2/pull/2092#issuecomment-1479788234) should improve the security boundaries. Still, it seems to be orthogonal to the Alerts setup. `notification-controller` can listen to events in any namespace anyway if the `cross-namespace references` flag is enabled. If this is inappropriate cluster admin can disable `cross-namespace references`. 

Are there any ongoing discussions about the threat model? I see that https://github.com/fluxcd/flux2/pull/2092 and https://github.com/fluxcd/flux2/pull/2093 were closed both to wait for https://github.com/kubernetes/enhancements/blob/master/keps/sig-auth/3766-referencegrant/README.md

xTrekStorex:
any update on this?

gecube:
Hi! Is it correct to say that this PR is still on hold and we are waiting?

makkes:
> Hi! Is it correct to say that this PR is still on hold and we are waiting?

Yes, pretty much. Current consensus is that we want to wait for the Kubernetes ReferenceGrant API to become GA which will definitely take a couple of Kubernetes releases to come.

MOZGIII:
This feature request seems irrelevant to `ReferenceGrant` API; what `ReferenceGrant` will affect is all of the cross-namespace references filtering - but that'd have to work at a different level. I see no point in holding on with this PR.
The current model is pretty secure - this PR does not make it weaker.

rinzler-17:
Hi, any update on this PR?

ojsef39:
Hey, any updates on this? Thanks in advance :)

dm3ch:
> There are security concerns related to this feature that were articulated in #71 (e.g. [#71 (comment)](https://github.com/fluxcd/notification-controller/issues/71#issuecomment-991049078)). Given the mentioned RFC has never been merged, we need to think this over.
> 
> In any case this will need to be held until notification-controller 1.0.0 has been released. Let's please take the time to completely discuss the threat model and potential mitigations here.

notification-controller passed 1.0.0 at Jul 2023.
KEP-3766 closed as not planned, so wouldn't be implemented any time soon.
What is currently holding this PR from being merged?

eloo-abi:
Hi, 
i was just running into the same limitation and then found this old topic here. 
What is the current here? Is there maybe already a solution? 
Or whats blocking this topic? 

Thanks

dm3ch:
As far as I understood this, flux team don't want to see wildcard functionality in mainline notification-controller as it brings questions how to implement effective boundaries in case of cross-NS logic especially with use-cases where cluster can have multiple flux installations.

In the meantime in k8s ecosystem there's still no a standard ways for operators/CRDs to implement such boundaries.

Please correct me if I'm wrong.

So from my POV the only way to get this functionality right now is to fork notification-controller or implement own custom notification controller or find a way to autogenerate notification manifests (for example with Kyverno)

stefanprodan:
I'm going to close this PR once we have a [ResourceSet](https://fluxoperator.dev/docs/resourcesets/app-definition/) example of how to sync the provider secret into every namespace and generare an Alert/Provider pair in every tenant namespace. The solution is very simple, requires no code changes, and it doesn't break Flux multi-tenancy nor does it bypasses Kubernetes RBAC.

eloo-abi:
@stefanprodan 

> sync the provider secret into every namespace and generare an Alert/Provider pair in every tenant namespace

Not sure if this is going to match the same usecase. 
IMHO this feels more like a "single namespace / tenant" use-case.. 

But if you want clusterwide notifications you would still miss the point with syncing alerts and providers into all namespaces. 

Syncing the Provider into each team namespace would also mean to sync the related secrets into the teams namespaces. 
Which is maybe not what we want to do. 

Maybe a dedicated CRD like "ClusterAlert" or so would be a better fit here.

PR #506, opened 2023-04-17
index receivers using webhook path as key

aryan9600:
Use `.status.webhookPath` as a key to index Receivers. Use this key while listing Receivers during the handling of a payload.
This enables us to list only the Receivers required instead of listing all the Receivers in the entire cluster.

### DIFF
--- a/internal/server/receiver_handler_test.go
+++ b/internal/server/receiver_handler_test.go
@@ -321,7 +321,7 @@ func Test_handlePayload(t *testing.T) {
 					Conditions:  []metav1.Condition{{Type: meta.StalledCondition, Status: metav1.ConditionFalse}},
 				},
 			},
-			expectedResponseCode: http.StatusNotFound,
+			expectedResponseCode: http.StatusServiceUnavailable,
 		},
 		{
 			name: "suspended receiver ignored",
@@ -337,7 +337,7 @@ func Test_handlePayload(t *testing.T) {
 					Conditions:  []metav1.Condition{{Type: meta.ReadyCondition, Status: metav1.ConditionTrue}},
 				},
 			},
-			expectedResponseCode: http.StatusNotFound,
+			expectedResponseCode: http.StatusServiceUnavailable,
 		},
 		{
 			name: "missing apiVersion in resource",
@@ -372,7 +372,7 @@ func Test_handlePayload(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedResponseCode: http.StatusBadRequest,
+			expectedResponseCode: http.StatusInternalServerError,
 		},
 		{
 			name: "resource by name not found",
@@ -406,7 +406,7 @@ func Test_handlePayload(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedResponseCode: http.StatusBadRequest,
+			expectedResponseCode: http.StatusInternalServerError,
 		},
 		{
 			name: "annotating resources by label match",
@@ -563,7 +563,7 @@ func Test_handlePayload(t *testing.T) {
 					"token": []byte("token"),
 				},
 			},
-			expectedResponseCode: http.StatusBadRequest,
+			expectedResponseCode: http.StatusInternalServerError,
 		},
 		{
 			name: "resource matchLabels is ignored if name is not *",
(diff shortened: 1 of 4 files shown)


makkes (commented):
Would you mind adding some more information about why this change is necessary to the PR description, please?

internal/server/receiver_handler_test.go
```diff
@@ -639,6 +639,10 @@ func Test_handlePayload(t *testing.T) {
 			if tt.receiver != nil {
 				builder.WithObjects(tt.receiver)
 			}
+			builder.WithIndex(&apiv1.Receiver{}, WebhookPathIndexKey, func(o client.Object) []string {
+				receiver := o.(*apiv1.Receiver)
+				return []string{receiver.Status.WebhookPath}
```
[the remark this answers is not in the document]
aryan9600: yes, i thought about this too, but then decided against it, because we'd need to define the index func in the `internal/server` package but then no actual code in the package would be using it (only a single test), whereas the main place of its use would be in `internal/controller`, so i find that a bit weird, semantically speaking.
but if you think its fine to have such a thing, then i don't really mind moving it.

stefanprodan (approved):
LGTM

Thanks @aryan9600 🏅

internal/server/receiver_handlers.go
```diff
@@ -71,8 +87,7 @@ func (s *ReceiverServer) handlePayload() func(w http.ResponseWriter, r *http.Req
 		receivers := make([]apiv1.Receiver, 0)
 		for _, receiver := range allReceivers.Items {
 			if !receiver.Spec.Suspend &&
```
hiddeco, line 89: Wondering if for suspended resources, we should throw a `http.StatusServiceUnavailable` ([HTTP 503](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/503)).

hiddeco: In addition, I think we expect a webhook path to be unique. Given this, I think we can instruct the `List` call to only return 1 result.

hiddeco: See https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#Limit

aryan9600: about returning a HTTP 503 response for suspended and/or unready resources, i'm not quite sure. according to the mdn docs:
>503 Service Unavailable server error response code indicates that the server is not ready to handle the request.
Common causes are a server that is down for maintenance or that is overloaded.

But a resource not being ready or suspended doesn't really map to the fact that the server is unable to handle the request.

aryan9600: maybe we could throw a `http.FailedDependency` ([HTTP 424](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/424)) instead, but according to mdn:
> Regular web servers will normally not return this status code. But some other protocols, like [WebDAV](https://developer.mozilla.org/en-US/docs/Glossary/WebDAV), can return it.

hiddeco: > Unavailable server error response code indicates that the server is not ready to handle the request.

It is not ready because "not ready" indicates it isn't (but eventually will be), and "suspend" is usually equal to "maintenance". Which is why I picked this.

aryan9600: okay, let's go with 503 since there is no other obvious answer 👍

hiddeco (approved):
Much less computing and more precise reporting about endpoint availability in minimal LOC of changes, absolute great work @aryan9600 :1st_place_medal:

internal/server/receiver_handlers.go
```diff
@@ -61,45 +76,48 @@ func (s *ReceiverServer) handlePayload() func(w http.ResponseWriter, r *http.Req
...
+			return
+		}
+
+		var withErrors bool
+		for _, resource := range receiver.Spec.Resources {
+			if err := s.requestReconciliation(ctx, logger, resource, receiver.Namespace); err != nil {
+				logger.Error(err, "unable to request reconciliation")
+				withErrors = true
 			}
 		}
```
hiddeco, line 123: One nit: I think `withErrors` below should actually return a HTTP 500. As the request can be made again without making a modification to the request.

Which isn't true for HTTP 400: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/400

hiddeco (approved):
Thanks for putting the icing on the :cake:

PR #508, opened 2023-04-21
receiver/v1: Remove deprecated `.status.url` field

darkowlzz:
Receiver `.status.url` was deprecated in v1beta2 API and replaced with `.status.webhookPath`. Remove it from v1 API and all references to it.


stefanprodan (approved):

makkes (approved):
Thanks @darkowlzz

PR #514, opened 2023-04-27
Allow customisation of the status-name.

bigkevmcd:
This allows customisation of the status name that's sent to Azure devops.

(I'm open to different approaches here, just trying to figure out a way).

Currently we send `fmt.Sprintf("%s/%s/%s", event.InvolvedObject.Kind, event.InvolvedObject.Name, uidParts[0])`

This is somewhat obscure, as it doesn't tie back to the originating cluster or resources, we had a customer try and send it in the Summary field for the notification, but this basically gets dropped on the floor by the Azure provider.


internal/notifier/azure_devops.go
```diff
@@ -172,3 +179,12 @@ func duplicateAzureDevOpsStatus(statuses *[]git.GitStatus, status *git.GitStatus
 
 	return false
 }
+
+func commitStatusGenre(event eventv1.Event) string {
+	summary, ok := event.Metadata["summary"]
+	if ok {
+		return fmt.Sprintf("%s:%s", genre, summary)
```
stefanprodan, line 186: Does Azure DevOps supports free-form text as genre? The summary can be 255 in length, is that supported?

stefanprodan: Doesn't say anything about max length, but Ok it seems that adding summary there could work.

stefanprodan (approved):
LGTM

Thanks @bigkevmcd

PR #518, opened 2023-04-30
Add config checksum from event metadata to rate limiting key

matheuscscp:
Builds on https://github.com/fluxcd/notification-controller/pull/517 and https://github.com/fluxcd/pkg/pull/544

Fixes https://github.com/fluxcd/helm-controller/issues/678 alongside https://github.com/fluxcd/helm-controller/pull/681 and https://github.com/fluxcd/pkg/pull/544

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -62,7 +62,7 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 			return
 		}
 
-		cleanupMetadata(event)
+		event.Metadata = cleanupMetadata(event)
 
 		ctx, cancel := context.WithTimeout(r.Context(), 15*time.Second)
 		defer cancel()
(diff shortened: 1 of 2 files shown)


hiddeco:
First wondered if the more generic approach to this would be to pass on the generation of the involved object, and take this into account. However, this would not cover for things which may produce different results due to to e.g. contents of referenced Secrets and/or ConfigMap changing (as could be the case for Helm values).

matheuscscp:
> First wondered if the more generic approach to this would be to pass on the generation of the involved object, and take this into account. However, this would not cover for things which may produce different results due to to e.g. contents of referenced Secrets and/or ConfigMap changing (as could be the case for Helm values).

But nice catch! We could also add the object generation anyways :thinking: Should I update this PR or open another one?

matheuscscp:
@hiddeco What are the next steps here now that https://github.com/fluxcd/pkg/pull/544 got merged?

hiddeco:
Update the other PRs to properly working with this, while the actual tagging of the event API package is waiting you can use a revision from main in the PRs.

matheuscscp:
Superseded by https://github.com/fluxcd/notification-controller/pull/530

Issue #511, opened 2023-04-25
Feature request: Add field to `Alert` CRD for adding metadata to the event


aristapimenta:
At the moment the only field that can be used for adding more context to an event is `.spec.summary`, which needs to be a string.

It would be very nice to have a first-class way to enhance the context of an event with a proper metadata dictionary field.

PR #519, opened 2023-05-01
Add event metadata field to Alert spec

matheuscscp:
Fixes https://github.com/fluxcd/notification-controller/issues/511

### DIFF
--- a/api/v1beta2/alert_types.go
+++ b/api/v1beta2/alert_types.go
@@ -50,6 +50,13 @@ type AlertSpec struct {
 	// +optional
 	InclusionList []string `json:"inclusionList,omitempty"`
 
+	// EventMetadata is an optional field for adding metadata to events emitted by the
+	// controller. Metadata fields added by the controller have priority over the fields
+	// added here, and the fields added here have priority over fields originally present
+	// in the event.
+	// +optional
+	EventMetadata map[string]string `json:"eventMetadata,omitempty"`
+
 	// ExclusionList specifies a list of Golang regular expressions
 	// to be used for excluding messages.
 	// +optional
--- a/docs/spec/v1beta2/alerts.md
+++ b/docs/spec/v1beta2/alerts.md
@@ -162,6 +162,35 @@ starting the controller with the `--no-cross-namespace-refs=true` flag.
 When this flag is set, alerts can only refer to event sources in the same namespace as the alert object,
 preventing tenants from subscribing to another tenant's events.
 
+### Event metadata
+
+`.spec.eventMetadata` is an optional field for adding metadata to events emitted by the
+controller. Metadata fields added by the controller have priority over the fields
+added here, and the fields added here have priority over fields originally present
+in the event.
+
+#### Example
+
+Add metadata fields to successful `HelmRelease` events:
+
+```yaml
+---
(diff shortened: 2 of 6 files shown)


internal/server/event_handlers.go
```diff
@@ -256,13 +256,15 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
...
-					notification.Metadata = map[string]string{
-						"summary": alert.Spec.Summary,
-					}
-				} else {
-					notification.Metadata["summary"] = alert.Spec.Summary
+				meta := notification.Metadata
+				if meta == nil {
+					meta = make(map[string]string)
+				}
+				for key, value := range alert.Spec.EventMetadata {
```
stefanprodan, line 263: This should happen even if summary is string empty

aristapimenta: I'm confused :thinking:

aristapimenta: oh, right!!! lemme fix it

aristapimenta: fixed :ok_hand:

stefanprodan: Thanks, looks good. We'll need to wait for GitHub to recover as now Actions are down.

docs/spec/v1beta2/alerts.md
```diff
@@ -162,6 +162,13 @@ starting the controller with the `--no-cross-namespace-refs=true` flag.
 When this flag is set, alerts can only refer to event sources in the same namespace as the alert object,
 preventing tenants from subscribing to another tenant's events.
 
+### Event metadata
+
+`.spec.eventMetadata` is an optional field for adding metadata to events emitted by the
+controller. Metadata fields added by the controller have priority over the fields
+added here, and the fields added here have priority over fields originally present
+in the event.
+
```
stefanprodan, line 171: Please add here an example section with metadata for:
```yaml
app.kubernetes.io/env : "production"
app.kubernetes.io/cluster: "my-cluster"
app.kubernetes.io/region: "us-east-1"
```

matheuscscp: done!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

api/v1beta2/alert_types.go
```diff
@@ -50,6 +50,13 @@ type AlertSpec struct {
 	// +optional
 	InclusionList []string `json:"inclusionList,omitempty"`
 
+	// EventMetadata is an optional field for adding metadata to events emitted by the
+	// controller. Metadata fields added by the controller have priority over the fields
+	// added here, and the fields added here have priority over fields originally present
+	// in the event.
```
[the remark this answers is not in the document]
hiddeco: After going through the PR, I have the same confusion.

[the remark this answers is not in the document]
matheuscscp: the controller mentioned here which has priority over everything else is notification-controller. when the event arrives at notification-controller, first the metadata is cleaned up, then the `eventMetadata` field from the `Alert` spec is added, and then finally the `summary` field from the `Alert` spec is added (the only one at the moment). the priority is given by this order.

i will file a PR to improve this explanation

PR #527, opened 2023-05-12
Add support for PagerDuty

MartiUK:
This PR adds support for PagerDuty.

There are some assumptions made in regards to what should trigger an incident, I'm happy to be corrected on this.

It would be nice to allow Alerts to override some Provider settings e.g. override the channel (integration key) to send alerts to different PagerDuty integrations without needed to create a Provider for each.

### DIFF
--- a/docs/spec/v1beta2/providers.md
+++ b/docs/spec/v1beta2/providers.md
@@ -118,6 +118,7 @@ The supported alerting providers are:
 | [Matrix](#matrix)                                       | `matrix`         |
 | [Microsoft Teams](#microsoft-teams)                     | `msteams`        |
 | [Opsgenie](#opsgenie)                                   | `opsgenie`       |
+| [PagerDuty](#pagerduty)                                 | `pagerduty`      |
 | [Prometheus Alertmanager](#prometheus-alertmanager)     | `alertmanager`   |
 | [Rocket](#rocket)                                       | `rocket`         |
 | [Sentry](#sentry)                                       | `sentry`         |
@@ -770,6 +771,64 @@ metadata:
   namespace: default
 stringData:
     token: xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
+```
+
+##### PagerDuty
+
+When `.spec.type` is set to `pagerduty`, the controller will send a payload for
+an [Event](events.md#event-structure) to the provided PagerDuty [Address](#address).
+
+The Event will be formatted into an [Event API v2](https://developer.pagerduty.com/api-reference/368ae3d938c9e-send-an-event-to-pager-duty) payload,
+triggering or resolving an incident depending on the event's `Severity`.
+
+The provider will also send [Change Events](https://developer.pagerduty.com/api-reference/95db350959c37-send-change-events-to-the-pager-duty-events-api)
+for `info` level `Severity`, which will be displayed in the PagerDuty service's timeline to track changes.
+
+This Provider type supports the configuration of a [proxy URL](#https-proxy)
+and [TLS certificates](#tls-certificates).
+
+The [Channel](#channel) is used to set the routing key to send the event to the appropriate integration.
+
+###### PagerDuty example
+
+To configure a Provider for Pagerduty, create a `pagerduty` Provider,
+set `address` to the integration URL and `channel` set to
+the integration key (also known as a routing key) for your [service](https://support.pagerduty.com/docs/services-and-integrations#create-a-generic-events-api-integration)
+or [event orchestration](https://support.pagerduty.com/docs/event-orchestration).
+
+When adding an integration for a service on PagerDuty, it is recommended to use `Events API v2` integration.
+
+**Note**: PagerDuty does not support Change Events when sent to global integrations, such as event orchestration.
+
+```yaml
+---
+apiVersion: notification.toolkit.fluxcd.io/v1beta2
+kind: Provider
+metadata:
+  name: pagerduty
+  namespace: default
+spec:
+  type: pagerduty
+  address: https://events.pagerduty.com
+  channel: <integrationKey>
+```
+If you are sending to a service integration, it is recommended to set your Alert to filter to
+only those sources you want to trigger an incident for that service. For example:
+
+```yaml
+---
+apiVersion: notification.toolkit.fluxcd.io/v1beta2
+kind: Alert
+metadata:
+  name: my-service-pagerduty
+  namespace: default
+spec:
+  providerRef:
+    name: pagerduty
+  eventSources:
+    - kind: HelmRelease
+      name: my-service
+      namespace: default
 ```
 
 ##### Prometheus Alertmanager
--- a/internal/notifier/pagerduty.go
+++ b/internal/notifier/pagerduty.go
@@ -1,0 +1,126 @@
+/*
+Copyright 2023 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package notifier
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+	"net/url"
+	"time"
+
+	"github.com/PagerDuty/go-pagerduty"
+
+	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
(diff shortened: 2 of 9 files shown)


internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,132 @@
...
+	if event.Severity == eventv1.EventSeverityError {
+		e.Action = "trigger"
+		e.Payload = &pagerduty.V2Payload{
+			Summary:   desc + ": " + name,
+			Source:    "Flux " + event.ReportingController,
+			Severity:  toPagerDutySeverity(event.Severity),
+			Timestamp: event.Timestamp.Format(time.RFC3339),
+			Component: event.InvolvedObject.Name,
+			Group:     event.InvolvedObject.Kind,
+			Details: map[string]interface{}{
```
stefanprodan, line 102: We need to add the event metadata to the details, the metadata can contain various infos such as source revision, cluster name, env, region, etc

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,132 @@
...
+		return nil, fmt.Errorf("invalid PagerDuty endpoint URL %s: '%w'", endpoint, err)
+	}
+	pHostname := pURL.Hostname()
+	if !slices.Contains(validEndpoints, pHostname) {
+		return nil, fmt.Errorf("invalid PagerDuty endpoint hostname %s: should be one of: %v", pHostname, validEndpoints)
+	}
+	opts := []pagerduty.ClientOptions{
+		pagerduty.WithV2EventsAPIEndpoint("https://" + pHostname),
+	}
+	// No token needed for our usage of the V2 events API
```
stefanprodan, line 55: Not sure I undestand how this works, anyone can spam PagerDuty if it knows the URL?

MartiUK: Yeah, the url is the same for everyone (except EU customers). The routing is done by the integration key. You can be rate limited, which I haven't handled yet.

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,130 @@
...
+	if event.Severity == eventv1.EventSeverityError {
+		e.Action = "trigger"
+		e.Payload = &pagerduty.V2Payload{
+			Summary:   desc + ": " + name,
+			Source:    "Flux " + event.ReportingController,
+			Severity:  toPagerDutySeverity(event.Severity),
+			Timestamp: event.Timestamp.Format(time.RFC3339),
+			Component: event.InvolvedObject.Name,
+			Group:     event.InvolvedObject.Kind,
+			Details:   event.Metadata,
```
stefanprodan, line 102: What about the message? You don't want to see the actual error in PagerDuty?

MartiUK: Yeah, you're right

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	// No token needed for our usage of the V2 events API
+	client := pagerduty.NewClient("", opts...)
+	return &PagerDuty{
+		Client:     client,
+		RoutingKey: routingKey,
+	}, nil
+}
+
+func (p *PagerDuty) Post(ctx context.Context, event eventv1.Event) error {
```
makkes, line 66: This is the central method here but there is no tests for it. I would suggest to write additional tests verifying proper functionality of this method. I'm not sure how the PagerDuty API works but maybe it's enough to spin up a test HTTP server ([net/http/httptest](https://pkg.go.dev/net/http/httptest)) and run the notifier against it.

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,178 @@
...
+
+type PagerDuty struct {
+	Client     *pagerduty.Client
+	RoutingKey string
+}
+
+func NewPagerDuty(endpoint string, routingKey string) (*PagerDuty, error) {
+	pURL, err := url.ParseRequestURI(endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("invalid PagerDuty endpoint URL %s: '%w'", endpoint, err)
```
makkes, line 49:
```suggestion
		return nil, fmt.Errorf("invalid PagerDuty endpoint URL %q: '%w'", endpoint, err)
```

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	pHostname := pURL.Hostname()
+	if !slices.Contains(validEndpoints, pHostname) {
+		return nil, fmt.Errorf("invalid PagerDuty endpoint hostname %s: should be one of: %v", pHostname, validEndpoints)
```
makkes, line 53:
```suggestion
		return nil, fmt.Errorf("invalid PagerDuty endpoint hostname %q: should be one of: %q", pHostname, validEndpoints)
```

internal/notifier/pagerduty.go
```diff
@@ -0,0 +1,178 @@
...
+	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/jpillora/backoff"
+	"k8s.io/utils/strings/slices"
+)
+
+var (
+	validEndpoints = []string{
+		"events.pagerduty.com",
+		"events.eu.pagerduty.com",
```
makkes, line 35: I'm not sure we'll have to validate these here. Whenever PagerDuty decides to change this list, we'll have to catch up and users won't be able to use the new endpoint address until a new Flux release has been published.

MartiUK: I wanted to ensure PagerDuty users that are using the EU servers (such as ourselves) were accounted for. I can remove the validation if you think isn't necessary.

MartiUK: I have removed the validation.

internal/notifier/pagerduty.go
MartiUK: This has been removed.

internal/notifier/pagerduty.go
MartiUK: Tests have been added

Whisper40:
Thanks for this PR @MartiUK !

MartiUK:
Hi @makkes, @stefanprodan is there anything else needed to progress this PR further?

internal/notifier/pagerduty_test.go
```diff
@@ -0,0 +1,227 @@
...
+	"testing"
+	"time"
+
+	"github.com/PagerDuty/go-pagerduty"
+	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/fluxcd/pkg/apis/meta"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
makkes, line 19:
```suggestion
	"github.com/PagerDuty/go-pagerduty"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
	corev1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"


	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
	"github.com/fluxcd/pkg/apis/meta"
```

makkes (commented):
Just a couple of nits, otherwise lgtm

makkes:
Please resolve the conflicts with `main` and squash the commits into one.

docs/spec/v1beta2/providers.md
```diff
@@ -875,6 +876,63 @@ stringData:
...
+  name: pagerduty
+  namespace: default
+spec:
+  type: pagerduty
+  address: https://events.pagerduty.com
+  channel: <integrationKey>
+```
+If you are sending to a service integration, it is recommended to set your Alert to filter to
+only those sources you want to trigger an incident for that service. For example:
+```yaml
```
stefanprodan, line 920:
```suggestion
If you are sending to a service integration, it is recommended to set your Alert to filter to
only those sources you want to trigger an incident for that service. For example:

```yaml
```

makkes (approved):
LGTM, thank you! 🥳

docs/spec/v1beta2/providers.md
```diff
@@ -124,6 +124,7 @@ The supported alerting providers are:
 | [Slack](#slack)                                         | `slack`          |
 | [Telegram](#telegram)                                   | `telegram`       |
 | [WebEx](#webex)                                         | `webex`          |
+| [PagerDuty](#pagerduty)                                 | `pagerduty`      |
```
stefanprodan, line 127: Please move PagerDuty right before Prometheus. Also in the section below with examples.

stefanprodan (commented):
This looks great, some tiny nits only from me.

makkes:
Your generated CRD manifests don't match the code. Please run `make manifests` and commit the changes. Also please squash your commits.

stefanprodan (approved):
LGTM

Thanks @MartiUK 🏅

PR #530, opened 2023-05-19
Include eventv1.MetaTokenKey on event rate limiting key calculation

matheuscscp:
Part of fixing https://github.com/fluxcd/helm-controller/issues/678

Supersedes https://github.com/fluxcd/notification-controller/pull/518

### DIFF
--- a/internal/server/event_server.go
+++ b/internal/server/event_server.go
@@ -192,9 +192,14 @@ func eventKeyFunc(r *http.Request) (string, error) {
 		event.Message,
 	}
 
-	revString, ok := event.Metadata[eventv1.MetaRevisionKey]
+	revision, ok := event.Metadata[eventv1.MetaRevisionKey]
 	if ok {
-		comps = append(comps, revString)
+		comps = append(comps, revision)
+	}
+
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/Azure/azure-event-hubs-go/v3 v3.5.0
 	github.com/containrrr/shoutrrr v0.7.1
 	github.com/fluxcd/notification-controller/api v1.0.0-rc.3
-	github.com/fluxcd/pkg/apis/event v0.4.1
+	github.com/fluxcd/pkg/apis/event v0.4.2-0.20230516071635-f4cf2dc1a8b0
```
hiddeco, line 14: We are awaiting a controller-runtime release to pair with an update to Kubernetes 1.27.x, which would probably be best to land before tagging this.

matheuscscp: When is that expected to happen?

hiddeco: Hopefully this week, as it is also blocking Kustomize v5 😬

hiddeco: This has seen a release now.

stefanprodan (approved):
LGTM

PR #531, opened 2023-05-22
Add `insecure-allow-http` flag to block HTTP traffic

aryan9600:
Add a new flag `insecure-allow-http` that blocks the controller from reaching any HTTP endpoints. Its set to true by default to avoid making a breaking change. If HTTP traffic is disabled and a `Provider` specifies an address with the `http` scheme, the reconciler errors out and uses the `InsecureConnectionsDisallowed` reason for the object's conditions.

Ref: https://github.com/fluxcd/flux2/tree/main/rfcs/0004-insecure-http


main.go
```diff
@@ -90,6 +91,7 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", 5*time.Minute, "Interval in which rate limit has effect.")
+	flag.BoolVar(&insecureAllowHTTP, "insecure-allow-http", true, "Allow the controller to make plain HTTP connections.")
```
stefanprodan, line 94:
```suggestion
	flag.BoolVar(&insecureAllowHTTP, "insecure-allow-http", true, "Allow the controller to forward events using  plain HTTP connections.")
```

internal/controllers/alert_controller_test.go
```diff
@@ -186,13 +187,33 @@ func TestAlertReconciler_EventHandler(t *testing.T) {
 	stopCh := make(chan struct{})
 	go eventServer.ListenAndServe(stopCh, eventMdlw, store)
 
-	rcvServer := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	// Run a TLS server, since HTTP traffic is disbaled for the ProviderReconciler.
```
stefanprodan, line 190:
```suggestion
	// Run a TLS server, since HTTP traffic is disabled for the ProviderReconciler.
```

PR #534, opened 2023-05-24
Prefix event key attributes with identifier

hiddeco:
This to prevent collisions between different event attributes.

For example when first an event with revision `foo` is received without a token, after which a token `foo` is received without a revision.

Based on a thought exchange in https://github.com/fluxcd/helm-controller/pull/692#pullrequestreview-1441751192.


matheuscscp (approved):
Thanks!!

PR #540, opened 2023-06-01
Refactor events and introduce v1beta3 API for Alert and Provider

darkowlzz:
Refer https://github.com/fluxcd/notification-controller/issues/626 for more details.

Based on a lot of observations from v1beta2 API discussion in https://github.com/fluxcd/notification-controller/pull/435, it became apparent that Alert and Provider can be static data objects without reconcilers and status subresource.

These changes are based on the experiments from last November, refer https://github.com/fluxcd/notification-controller/commits/refactor-event-handler, updated to include all the latest changes.

- Add Alert and Provider v1beta3 APIs which make them static API objects removing the need for any reconciler.
- Remove the old Alert and Provider reconcilers.
- Refactor event handler, splitting it into smaller functions with unit test and integration test of the event server. Also, the event handlers emit k8s events on the respective Alert object for the events.
- Add new Alert and Provider reconcilers to help in migration to static API objects. This is needed to remove the finalizer set by the previous reconcilers. Stale finalizers would interfere with object deletion. Since the new API don't have status, the status from the old objects are automatically dropped when the new API is used. These are temporary migration reconcilers and will be removed in a subsequent release. (__NOTE:__ These new reconcilers are added separately, not on top of the previous reconcilers. Reading commit-by-commit would make it easier to understand the changes.)
- Add new v1beta3 spec docs.

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -202,9 +202,9 @@ func (s *EventServer) dispatchNotification(ctx context.Context, event *eventv1.E
 	}
 
 	go func(n notifier.Interface, e eventv1.Event) {
-		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		pctx, cancel := context.WithTimeout(context.Background(), timeout)
 		defer cancel()
-		if err := n.Post(ctx, e); err != nil {
+		if err := n.Post(pctx, e); err != nil {
 			maskedErrStr, maskErr := masktoken.MaskTokenFromString(err.Error(), token)
 			if maskErr != nil {
 				err = maskErr


matheuscscp (approved):
first pass LGTM

internal/controller/provider_controller.go
```diff
@@ -18,301 +18,79 @@ package controller
...
-	// Initialize the runtime patcher with the current version of the object.
-	patcher := patch.NewSerialPatcher(obj, r.Client)
-
-	defer func() {
-		// Patch finalizers, status and conditions.
-		if err := r.patch(ctx, obj, patcher); err != nil {
-			retErr = kerrors.NewAggregate([]error{retErr, err})
-		}
-
-		// Record Prometheus metrics.
```
stefanprodan, line 102: Don’t we want to preserve these metrics that people use to determine when resources are being suspended and/or deleted?

api/v1beta3/alert_types.go
```diff
@@ -0,0 +1,102 @@
...
+// +kubebuilder:storageversion
+// +kubebuilder:object:root=true
+// +kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp",description=""
+
+// Alert is the Schema for the alerts API
+type Alert struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec AlertSpec `json:"spec,omitempty"`
```
stefanprodan, line 88: Why not set status here with ready=true as the default? Without it, the Flux CLI and kstatus will get confused I think.

hiddeco: Based on https://github.com/kubernetes-sigs/cli-utils/blob/master/pkg/kstatus/status/status.go#L130-L139, I think this should work just fine with kstatus.

stefanprodan:
@darkowlzz I'm for dropping `caFile` support in v1beta3, please go ahead and delete the code and related docs if any.

stefanprodan (commented):
Please mark the v1beta2 APIs as deprecated.

From [docs](https://fluxcd.io/flux/releases/controllers/#beta-version):
>A beta version API becomes deprecated once a subsequent beta or stable API version is released. A deprecated beta version is subject to removal after a six-months period.

stefanprodan (approved):
LGTM

Thanks @darkowlzz 🥇 

PS. Please create a PR in website to move the API docs to `v1beta3` and update all docs to the new version.

stefanprodan:
@darkowlzz please integrate #639 and make the correction in docs to
```
[Bitbucket Server/Data Center](#bitbucket-serverdata-center)
```

stefanprodan (approved):

PR #558, opened 2023-06-20
Make Gitea tests independent of 3rd-party service

makkes:
try.gitea.io, used before this change for testing the Gitea notifier, has been provisioned with a bug that broke the tests. Now, the tests are run against a mock HTTP server mimicking parts of Gitea's API so that the tests don't rely on that 3rd-pary service, anymore.


stefanprodan (approved):
LGTM

Thanks @makkes 🏅

PR #584, opened 2023-07-24
Handle delete before adding finalizer

darkowlzz:
In `Reconcile()` methods, move the object deletion above add finalizer. Finalizers can't be set when an object is being deleted.

Found by @hiddeco while working on helm-controller. Refer https://github.com/kubernetes-sigs/cluster-api/commit/cef1cf1d9312088a785c520f49970b1c609ab9ad .


hiddeco (approved):

makkes (approved):
@darkowlzz Good spotting. Is there a way we can add a regression test for this case?

makkes (approved):

PR #585, opened 2023-07-30
change key type to []byte in provider spec docs

### LINKED ISSUE #581: Docs - v1beta2 - Providers - Key type
There is a type issue in the example provided for Generic Webhook with HMAC. 

https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta2/providers.md#generic-webhook-with-hmac

The function verifySignature takes a key of type byte array.
```go
func verifySignature(signature string, payload, key []byte) error
```

A string value is provided to the function when it gets called in the `handleRequest` function
```go
// Verify signature using the same token as the Secret referenced in
// Provider
key := "<token>"
if err := verifySignature(r.Header.Get("X-Signature"), body, key); err != nil {
    http.Error(w, fmt.Sprintf("failed to verify HMAC signature: %s", err.Error()), http.StatusBadRequest)
    return
}
```

michaelsatish:
fixes #581

### DIFF
--- a/docs/spec/v1beta2/providers.md
+++ b/docs/spec/v1beta2/providers.md
@@ -268,7 +268,7 @@ func handleRequest(w http.ResponseWriter, r *http.Request) {
 
 	// Verify signature using the same token as the Secret referenced in
 	// Provider
-	key := "<token>"
+	key := []byte("<token>")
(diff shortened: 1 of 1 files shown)


makkes (approved):
thanks @michaelsatish

PR #592, opened 2023-08-02
feat: add DataDog notification provider

mrparkers:
This PR adds support for DataDog as a notification provider.

This provider publishes events that look like this:
image

...given a provider config that looks like this:

```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta2
kind: Provider
metadata:
  name: datadog
  namespace: flux-system
spec:
  secretRef:
    name: datadog-secret
  type: datadog
  address: https://api.datadoghq.com
  channel: my-k8s-cluster-name
```

...and a secret that looks like this:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: datadog-secret
  namespace: flux-system
stringData:
  token: "datadog-api-key"
  headers: |
    my-custom-tag: tag-value
```

...and an alert that looks like this:

```yaml
kind: Alert
metadata:
  name: datadog-info
  namespace: flux-system
spec:
  eventSeverity: info
  eventSources:
    - kind: Kustomization
      name: "*"
    - kind: HelmRelease
      name: "*"
    - kind: GitRepository
      name: "*"
  providerRef:
    name: datadog
```

I made a few assumptions along the way based on the way other events / tags look within the DataDog ecosystem.  I tried to leave comments in the code that explains the rationale behind each of these.

### DIFF
--- a/docs/spec/v1beta2/providers.md
+++ b/docs/spec/v1beta2/providers.md
@@ -109,6 +109,7 @@ The supported alerting providers are:
 | [Generic webhook](#generic-webhook)                     | `generic`        |
 | [Generic webhook with HMAC](#generic-webhook-with-hmac) | `generic-hmac`   |
 | [Azure Event Hub](#azure-event-hub)                     | `azureeventhub`  |
+| [DataDog](#datadog)                                     | `datadog`        |
 | [Discord](#discord)                                     | `discord`        |
 | [GitHub dispatch](#github-dispatch)                     | `githubdispatch` |
 | [Google Chat](#google-chat)                             | `googlechat`     |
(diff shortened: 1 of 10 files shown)


stefanprodan:
Please use `eventMetadata` to set the custom tags instead of things in the secret along with the standard ones like revision and summary. The cluster name, as specified in docs should come from eventMetadata, the env can be set from the channel.

internal/notifier/datadog.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	// set the cluster name to the value of the summary field if it exists
+	// otherwise, set it to the value of the channel
+	if v, ok := event.Metadata["summary"]; ok {
+		tags = append(tags, fmt.Sprintf("kube_cluster_name:%s", v))
+		delete(event.Metadata, "summary")
+	} else {
+		tags = append(tags, fmt.Sprintf("kube_cluster_name:%s", d.env))
+	}
```
stefanprodan, line 157: I would not do this, if people want `kube_cluster_name` they should set in the Alert e.g.

```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta2
kind: Alert
metadata:
  name: <name>
spec:
  eventMetadata:
    kube_cluster_name: "my-cluster"
    kube_cluster_region: "my-region"
```

Let's use channel for env, all the rest should come from the Alert.

mrparkers:
Okay, I think that make sense.  I've removed the code that derives `kube_cluster_name` from the `summary` field of the event metadata.

docs/spec/v1beta2/providers.md
```diff
@@ -405,6 +406,51 @@ stringData:
...
 
+##### DataDog
+
+When `.spec.type` is set to `datadog`, the controller will send a payload for
+an [Event](events.md#event-structure) to the provided DataDog API [Address](#address).
+
+The Event will be formatted into a [DataDog Event](https://docs.datadoghq.com/api/latest/events/#post-an-event) and sent to the
+API endpoint of the provided DataDog [Address](#address).
+
+The Provider's [Channel](#channel) is used to set the `env` and `kube_cluster_name` tags
```
somtochiama, line 417: > and kube_cluster_name tags

I guess this is no longer the case after your latest changes. Only the env tag is now set.

internal/notifier/datadog.go
```diff
@@ -0,0 +1,176 @@
...
+// channel: The name of the Kubernetes cluster that is producing these events, mapped to the `env` tag in DataDog.
+// headers: A map of extra tags to add to the event
+func NewDataDog(address string, proxyUrl string, certPool *x509.CertPool, token string, channel string) (*DataDog, error) {
+	conf := datadog.NewConfiguration()
+
+	if token == "" {
+		return nil, fmt.Errorf("token cannot be empty")
+	}
+
+	if channel == "" {
```
somtochiama, line 53: Should the channel be required? It seems that the Datadog API doesn't require tags

docs/spec/v1beta2/providers.md
mrparkers: Thanks, I'll fix this.

internal/notifier/datadog.go
mrparkers: The `env` tag is usually used all over datadog (for metrics and logs) in order to enable [unified service tagging](https://docs.datadoghq.com/getting_started/tagging/unified_service_tagging), but you're right that it isn't a strict requirement on DataDog's side.

I understand your point about not necessarily needing to make this a requirement.  I made this decision before I realized that the event metadata via the `Alert` can be used by the user to add any tags they want.  So I'm okay with removing this requirement and expecting users to use event metadata to add this tag.

somtochiama: I think we can still have the `channel` set to `env` tag if it is specied. 
It just wouldn't be required for those who don't care about the tag.

docs/spec/v1beta2/providers.md
somtochiama: Thanks! and I just noticed there's also a comment for ` toDataDogTags` that also mentions the `kube_cluster_name` tag.

internal/notifier/datadog.go
mrparkers: I thought about it some more, and I actually think it's better to just not use `channel` at all, and expect users to set tags like `env` via the event metadata.  Is that approach alright?

cep21:
Thanks for this change! We use datadog and this will be very helpful.

stefanprodan:
@mrparkers can you please squash the commits and rebase with upstream main.

mrparkers:
Done!  Thanks again for the review!

stefanprodan (approved):
LGTM

Thanks @mrparkers 🏅 

PS. We'll need another rebase with upstream main to merge it.

mrparkers:
Hi @stefanprodan, I just rebased with main again.  Let me know if there's anything else you need from me 😄

PR #602, opened 2023-08-15
Fix Providers doc refer to non-existent 'events.md'

kingdonb:
Another commit to go with:

* https://github.com/fluxcd/source-controller/pull/1200

### DIFF
--- a/docs/spec/v1beta2/providers.md
+++ b/docs/spec/v1beta2/providers.md
@@ -144,7 +144,7 @@ The supported providers for [Git commit status updates](#git-commit-status-updat
 When `.spec.type` is set to `generic`, the controller will send an HTTP POST
 request to the provided [Address](#address).
 
-The body of the request is a [JSON `Event` object](events.md#event-structure),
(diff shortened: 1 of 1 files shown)


kingdonb:
There is no `events.md` - this is the result of anchor/404 testing being done on fluxcd/website#1573 and associated PRs

Please review and merge at will

stefanprodan:
Pinging @makkes

makkes:
Sorry for the confusion. We had to change the docs structure in #582 so that all API type pages are pluralized. The import-flux2-assets.sh script  in the `main` branch of the website repo will fail with the new structure until fluxcd/website#1577 is merged from which point it will pass again. Since we're not building the docs from the n-c `main` branch but only from a released version the docs structure change onf #582 has no impact on the current docs site.

I think we can close this PR and wait until fluxcd/website#1577 has been merged. Then, the anchor check should not fail, anymore.

kingdonb:
OK, 👍 closing

Issue #470, opened 2023-02-02
Add NATS as a notifier


peedrr:
NATS is a perfect candidate for a notifier; it is a cloud-native messaging service, it is CNCF incubating and it is written in Go. They have a fully-featured [Go client](https://github.com/nats-io/nats.go) so I believe integration should be fairly straightforward.

PR #651, opened 2023-11-26
Add Provider for NATS Subject

SkullKidCode:
Fixes https://github.com/fluxcd/notification-controller/issues/470

Adds support to send events to a NATS subject. Supports authentication through Username/Password as well.

Due the similarities in architecture with Google Pub/Sub, I drew inspiration for the implementation and testing of this change from that provider.

### DIFF
--- a/docs/spec/v1beta3/providers.md
+++ b/docs/spec/v1beta3/providers.md
@@ -107,6 +107,7 @@ The supported alerting providers are:
 | [Slack](#slack)                                         | `slack`          |
 | [Telegram](#telegram)                                   | `telegram`       |
 | [WebEx](#webex)                                         | `webex`          |
+| [NATS](#nats)                                           | `nats`           |
 
 The supported providers for [Git commit status updates](#git-commit-status-updates) are:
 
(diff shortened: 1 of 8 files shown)


SkullKidCode:
Updated to add the new provider to `v1beta3` instead since I see it has just been added :D

docs/spec/v1beta3/providers.md
```diff
@@ -973,6 +974,50 @@ stringData:
   token: <bot-token>
 ```
 
+##### NATS
+
+When `.spec.type` is set to `nats`, the controller will publish the payload of
+an [Event](events.md#event-structure) on the NATS Subject provided in the
```
stefanprodan, line 980:
```suggestion
an [Event](events.md#event-structure) on the [NATS Subject](https://docs.nats.io/nats-concepts/subjects) provided in the
```

stefanprodan: Let's add a link to the NATS Subject docs please.

stefanprodan (approved):
LGTM

Thanks @SkullKidCode 🏅

PR #683, opened 2023-12-12
Add .spec.interval in v1beta3 Provider

darkowlzz:
For backwards compatibility with v1beta2 Provider, add .spec.interval with deprecation marker.
Interval was an optional field in v1beta2 with internal default. Only the users who had explicitly set intervals would have encountered this incompatibility error after upgrade.

Fixes https://github.com/fluxcd/flux2/issues/4471.


hiddeco (approved):

api/v1beta3/provider_types.go
```diff
@@ -61,6 +61,15 @@ type ProviderSpec struct {
 	// +required
 	Type string `json:"type"`
 
+	// Interval at which to reconcile the Provider with its Secret references.
+	// Deprecated with static Provider.
```
stefanprodan, line 65:
```suggestion
	// Deprecated and not used in v1beta3.
```

stefanprodan (approved):

Issue #650, opened 2023-11-23
The token field should be excluded from the alter payload


stefanprodan:
In https://github.com/fluxcd/notification-controller/pull/530 we've added a token field to be used for rate limiting. This field is used internally and should've be part of the payload sent to alert providers.

<img width="672" alt="Screenshot 2023-11-23 at 09 59 41" src="https://github.com/fluxcd/notification-controller/assets/3797675/60ef5a37-68e7-4c5f-b59f-d3f8ad48ba19">

PR #686, opened 2023-12-13
Exclude eventv1.MetaTokenKey from event metadata

darkowlzz:
`eventv1.MetaTokenKey` is required to be considered in rate limiting but it is only for internal use by flux components and should not be sent to alert providers. Remove `eventv1.MetaTokenKey` from the metadata of event before processing it for matching alerts.
Fixes https://github.com/fluxcd/notification-controller/issues/650

hc-alert-exclude-token

NOTE: Initially, I added it in `cleanupMetadata()` and found out during manual testing that it removes the metadata too early which excludes it from rate limiter. Also tried adding it in `enhanceEventWithAlertMetadata()` but realized that this is too late. The same metadata will be removed for every single alert the event matches with. Removing the internal metadata right at the beginning of the event handler seems to be most appropriate.

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -58,6 +58,9 @@ func (s *EventServer) handleEvent() func(w http.ResponseWriter, r *http.Request)
 		ctx, cancel := context.WithTimeout(r.Context(), 15*time.Second)
 		defer cancel()
 
+		// Remove any internal metadata before further processing the event.
+		excludeInternalMetadata(event)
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @darkowlzz 🏅

hiddeco (approved):
This is in heavy need of more functional tests, but for now it looks good to me. Thanks @darkowlzz :bow:

Issue #700, opened 2023-12-29
Panic while using bitbucketserver provider for git commit status


gdasson:
### Describe the bug
The support for Bitbucket Server/Datacenter as a provider for git commit status was introduced by this [PR](https://github.com/fluxcd/notification-controller/pull/639)

Unfortunately, while testing the code using self signed certificates, I noticed a bug in the code that's causing a panic: runtime error: invalid memory address or nil pointer dereference error

### To Reproduce:
1. Enable bitbucket server provider for git commit status like below:
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta3
kind: Provider
metadata:
  name: server-commit-status-provider
  namespace: flux-system
spec:
  address: https://test.internal.server:8443/scm/foo/bar.git
  secretRef:
    name: token
  type: bitbucketserver
```

2. The certificate for the server is self signed, but certSecretRef was not provided

### Current Actual behavior:
Notification controller dies with a panic error:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x1659578]

goroutine 236 [running]:
github.com/fluxcd/notification-controller/internal/notifier.BitbucketServer.duplicateBitbucketServerStatus({{0x40007d3766, 0x5}, {0x40007d376c, 0xd}, {0x40006db680, 0x24}, {0x40007d3740, 0x3d}, {0x400071cb70, 0x21}, ...}, ...)
	github.com/fluxcd/notification-controller/internal/notifier/bitbucketserver.go:195 +0x258
github.com/fluxcd/notification-controller/internal/notifier.BitbucketServer.Post({{0x40007d3766, 0x5}, {0x40007d376c, 0xd}, {0x40006db680, 0x24}, {0x40007d3740, 0x3d}, {0x400071cb70, 0x21}, ...}, ...)
	github.com/fluxcd/notification-controller/internal/notifier/bitbucketserver.go:155 +0x3a8
github.com/fluxcd/notification-controller/internal/server.(*EventServer).dispatchNotification.func1({0x2008500, 0x4000523e60}, {{{0x40007c6870, 0xd}, {0x40007c6880, 0xb}, {0x40007be588, 0x18}, {0x40006daf60, 0x24}, ...}, ...})
	github.com/fluxcd/notification-controller/internal/server/event_handlers.go:210 +0xc0
[...]

PR #701, opened 2023-12-30
Make bitbucket server/data center git commit status functionality more robust

gdasson:
This is an extension to this initial draft [PR](https://github.com/fluxcd/notification-controller/pull/639)

This PR offers the below:
* Fixes this  [bug](https://github.com/fluxcd/notification-controller/issues/700)
* Adds more bitbucket server states: 
   * **Context**: Currently, the code supports only 2 bitbucket server build status states `SUCCESSFUL` or `FAILED` based on event severity types `info` or `error` respectively. 
       For some `info` events, the bitbucket server build status is incorrectly being set as `SUCCESSFUL`. For example:
       * an info event with event reason `DependencyNotReady` should ideally be set as `UNKNOWN`
       * an info event with event reason `Progressing` (for ex: Health checks still haven't passed),should ideally be set as `INPROGRESS`
       This PR adds these functionalities.

Reference: This REST API [document](https://developer.atlassian.com/server/bitbucket/rest/v815/api-group-builds-and-deployments/#api-api-latest-projects-projectkey-repos-repositoryslug-commits-commitid-builds-post) shows the different Bitbucket server build status states available for use.

Screenshot to show the different states on Bitbucket Server UI
Screenshot 2023-12-29 at 10 38 16 PM


stefanprodan (commented):
Hey @gdasson thanks for this, but we can't make this controller depend on another. I propose you open a new PR with just the bug fix. To handle the reasons, first we need to move them to `fluxcd/pkg` then we can use those here without adding cross-controller dependencies.

gdasson:
@stefanprodan thanks for the review and feedback. What is the process to add constants to fluxcd/pkg? Should I initiate a proposal conversation on flux-contributors slack channel? Or just submit a PR on fluxcd/pkg and kustomize-controller describing the proposal of movings constants to pkg? Please let me know. Thx.

stefanprodan:
@gdasson can you please open a separate PR to fix the bug so we can include it in the next patch release? The handling of status conditions can't go into a patch release and these should cover all Git providers not only BitBucket.

Issue #702, opened 2024-01-05
Grafana annotation for Kustomization does not show revision hash


radenui:
Hi !

When defining an alert on Kustomizations using Grafana provider, the field `revision` is not displayed correctly in grafana.

Example:
For a revision like `main@sha1:d78e2d410d0944b0cf15f1168e3167d71eb7c27f`, the tag set in Grafana API has the format `revision:<revisionvalue>` and grafana seems to split after the second `:` character.

It all happens in [this code](https://github.com/fluxcd/notification-controller/blob/main/internal/notifier/grafana.go#L74).

The resulting tag in grafana is rather unusable: `revision:main@sha1`.

I would suggest to either: 

- change the separator between `shaXX` and the actual hash: I imagine a replacement of all `:` characters in grafana provider as this is not specific to `revision` field
- just remove `shaX:` from `revision` field (in grafana provider only I guess, at it works fine with other providers)

stefanprodan:
This broke when we introduced the new revision format, previously it didn't include the hash type and `:`  character.

radenui:
Most probably, yes.
But the problem would be the same for custom eventMetadata, so I believe a local fix (replacement of `:` character) in grafana provider would be the best choice.

radenui:
Something like:
```go
for k, v := range event.Metadata {
  key := strings.ReplaceAll(k, ":", "-")
  value := strings.ReplaceAll(v, ":", "-")
  sfields = append(sfields, fmt.Sprintf("%s: %s", key, value))
}
```

PR #703, opened 2024-01-05
fix(grafana-provider): replace ":" character in eventMetadata

radenui:
Fix proposal for issue #702


stefanprodan (approved):
LGTM

Thanks @radenui

Issue #589, opened 2023-07-31
Alerts/Github - is it possible to separate add additional metadata?


audunsolemdal:
So my goal is to run a github action step to check if fluxcd reconciliation for a kustomization towards a  specificcluster was successful. There are three different kustomizations in three different clusters which may post `reconcilation successful` status to my commit on github. The kustomizations have the same exact name  It does not seem like modifying the `eventMetadata` or `summary` fields in the Alert resource adds any extra metadata to the github status.

I guess a workaround is to name the kustomizations `tstsp1-github-info-${cluster}` rather than just `tstsp1-github-info`, but I am wondering if there are other methods to inject this metadata into the Github status? The `description` or `context` fields would be sufficient to modify in my use case.

Say I have a repo with the following structure

orgX/wl-tstsp1:

```
fluxcd/
  dev/
    deployment.yaml
  test/
    deployment.yaml
  prod/
    deployment.yaml
```

Each folder is connected to the namespace called `tstsp1` in three different cluster (dev/test/prod)
From each of the clusters there is a alert updating the github status

```yaml
---
apiVersion: notification.toolkit.fluxcd.io/v1beta2
kind: Provider
metadata:
  name: tstsp1-github
  namespace: flux-system
spec:
  type: github
  address: https://github.com/orgX/wl-tstsp1
  secretRef:
    name: gh-flux-pat
---
apiVersion: notification.toolkit.fluxcd.io/v1beta2
kind: Alert
metadata:
  name: "tstsp1-github-info"
  namespace: flux-system
spec:
  providerRef:
    name: tstsp1-github
  eventSeverity: info
  summary: "dev01"
  eventMetadata:
    cluster: dev01
  eventSources:
    - kind: Kustomization
      name: '*'
      namespace: tstsp1
```

I end up with a status looking somewhat like the below:

```
gh api /repos/orgX/wl-tstsp1/statuses/88efe1eeb3e2ad9724acc6ff7c06324f6e613a66
[
  {
    "url": "https://api.github.com/repos/orgX/wl-tstsp1/statuses/88efe1eeb3e2ad9724acc6ff7c06324f6e613a66",
    "avatar_url": "https://avatars.githubusercontent.com/u/18458716?v=4",
    "id": 24342874371,
    "node_id": "SC_kwDOHpAnes8AAAAFqvLJAw",
    "state": "success",
    "description": "reconciliation succeeded",
    "target_url": null,
    "context": "kustomization/tstsp1/025ecd77",
[...]

PR #705, opened 2024-01-08
feat: add function to create a more speaking id for commit status

sebbrandt87:
Add a function to get a more detailed insight into the commit status of e.g. GitLab.
So that you do not need to lookup the first part of the Providers UIDs at all your Flux powered Kubernetes clusters.


internal/notifier/util.go
```diff
@@ -59,10 +59,29 @@ func formatNameAndDescription(event eventv1.Event) (string, string) {
...
+// createMetadataLabelsSuffix returns a concated string depending on app.kubernetes.io/,
+// prefixed common labels https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/
+func createMetadataLabelsSuffix(event eventv1.Event) string {
+	metadataLabels := make([]string, 0)
+	prefixedMetadataLabelsSuffix := ""
+	for labelKey, labelValue := range event.Metadata {
+		if strings.Contains(labelKey, "app.kubernetes.io/") {
+			metadataLabels = append(metadataLabels, labelValue)
+		}
+	}
```
makkes, line 76: This code doesn't assure a certain ordering in how the final string is constructed. You will also notice this when running the tests multiple times as they will fail occasionally.

I don't think we should implement such a non-deterministic behaviour as people may want to parse the string.

internal/notifier/util.go
```diff
@@ -59,10 +59,29 @@ func formatNameAndDescription(event eventv1.Event) (string, string) {
 // involved object kind and name.
 func generateCommitStatusID(providerUID string, event eventv1.Event) string {
 	uidParts := strings.Split(providerUID, "-")
-	id := fmt.Sprintf("%s/%s/%s", event.InvolvedObject.Kind, event.InvolvedObject.Name, uidParts[0])
+	metadataLabelsSuffix := createMetadataLabelsSuffix(event)
```
makkes, line 62: This has the potential to result in a very long string which is undesirable. We need to think on a way to constrain this to a certain length.

makkes (commented):
This change needs to be accompanied by documentation. I'm also not sure how much we're bought into the current format as people might already be relying on it as it is right now. @stefanprodan?

internal/notifier/util.go
sebbrandt87: sure, I am going to implement an string sorted keys construction.

internal/notifier/util.go
sebbrandt87: true, it might be an option to change eventMetadata keys accepted for building an extended commit status ID? @makkes

docs/spec/v1beta3/providers.md
```diff
@@ -1474,6 +1474,34 @@ spec:
       name: flux-system
 ```
 
+##### Construction of Git Commit Status ID
+
+The commit status ID is constructed through the `eventSources.kind`, `eventSources.name` and the
```
makkes, line 1479:
```suggestion
By default the commit status ID is constructed through the `eventSources.kind`, `eventSources.name` and the 
```

docs/spec/v1beta3/providers.md
```diff
@@ -1474,6 +1474,34 @@ spec:
       name: flux-system
 ```
 
+##### Construction of Git Commit Status ID
```
makkes, line 1477: We should explain here what the commit status ID actually is and how it can be leveraged.

docs/spec/v1beta3/providers.md
```diff
@@ -1474,6 +1474,34 @@ spec:
...
+
+The commit status ID is constructed through the `eventSources.kind`, `eventSources.name` and the 
+[`UID`](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#uids) of the `Provider`.
+
+To generate a speakable prefix, instead of relying on the `UID` of the `Provider`, `eventMetadata` can be set.
```
makkes, line 1482:
```suggestion
To generate a descriptive prefix, instead of relying on the `UID` of the `Provider`, `eventMetadata` can be set.
```

docs/spec/v1beta3/providers.md
```diff
@@ -1474,6 +1474,34 @@ spec:
...
+metadata:
+  name: github-status
+  namespace: flux-system
+spec:
+  providerRef:
+    name: github-status
+  eventMetadata:
+    app.kubernetes.io/env: "production"
+    app.kubernetes.io/cluster: "my-cluster"
+    app.kubernetes.io/region: "us-east-1"
```
makkes, line 1496: I'm not sure if the `app.kubernetes.io` is supposed to be used here. My interpretation has been that it's exclusive to well-known labels defined by Kubernetes.

sebbrandt87: I would count it to be okay as it is already in the [alerts.md](https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta3/alerts.md#example-1) mentioned

makkes: 👌🏻

internal/notifier/util.go
sebbrandt87: any suggestion how to proceed @makkes @stefanprodan ?

docs/spec/v1beta3/providers.md
sebbrandt87: Is it might be enough to link to GitHub and GitLab related API endpoints - as this documentation stays up2date? @makkes 

GitHub: https://docs.github.com/en/rest/commits/statuses?apiVersion=2022-11-28
GitLab: https://docs.gitlab.com/ee/api/commits.html#set-the-pipeline-status-of-a-commit

matheuscscp:
I'd prefer to make this PR superseded by this approach: https://github.com/fluxcd/notification-controller/issues/589#issuecomment-2688342204

@sebbrandt87 are you interested in contributing this feature? We could work together to get it done quicker

sebbrandt87:
> I'd prefer to make this PR superseded by this approach: [#589 (comment)](https://github.com/fluxcd/notification-controller/issues/589#issuecomment-2688342204)
> 
> @sebbrandt87 are you interested in contributing this feature? We could work together to get it done quicker

Hey @matheuscscp,
I think I understand your set approach but I don't see yet why we have to combine this here already. :)

I hope to find time to get this long running but still needed feature done. :)

matheuscscp:
Superseded by #1068

PR #722, opened 2024-01-30
Fix BitBucket status update panic

stefanprodan:
Fix: #700

Ported from: #701


gdasson:
Thanks for picking this up @stefanprodan. I apologize I have been busy these days with some personal stuff and couldn't spare time to work on it!

Issue #125, opened 2021-01-19
Spesifying notification Provider's namespace in an alert


ambis:
Running 0.6.1

Regarding this section:
https://github.com/fluxcd/notification-controller/blob/main/config/crd/bases/notification.toolkit.fluxcd.io_alerts.yaml#L92

It only allows you to specify the name of the alert Provider, in my case discord. But it does not allow to reference it form another namespace (monitoring in this case).

The intent I believe, in a single tenant cluster, is to have only one Provider, which sits nicely in the `flux-system` namespace.

But it is impossible now to create an alert in a namespace where there are eg. HelmReleases you want to watch (like monitoring):

```
---
apiVersion: notification.toolkit.fluxcd.io/v1beta1
kind: Alert
metadata:
  name: discord-alert-monitoring
  namespace: monitoring
spec:
  providerRef:
    name: discord
    namespace: flux-system   #  <<---- THIS IS NOT SUPPORTED
  eventSeverity: info
  eventSources:
    - kind: HelmRelease
      name: '*'
      #namespace: monitoring  # <<--- This is what I use now, when this resource is defined in the flux-system namespace, not monitoring
```

Is the intention of running all GOTK components in the `flux-system` namespace, including `Alerts`?

This issue https://github.com/fluxcd/notification-controller/issues/71 would solve my problems, since I could subscribe to all (single tenant) cluster events at once, without having to worry about setting up separate alerts per namespace.

phillebaba:
Is this something that we should consider implementing? It will break existing multi tenant setups if we switch over to allowing cross namespace references.

hrvatskibogmars:
👍

benfairless:
I personally would like the suggested behaviour. It would allow tenants to set up their own alerts using an existing integration.

sambonbonne:
Something like this would be useful for one of my clients: we have one provider for a lot of alerts and we want alerts to be created in same namespaces as Helm releases so each namespace would be more "self-contained".

rsafonseca:
I've opened a PR that solves this issue and a couple others :)

PR #743, opened 2024-02-15
Add shareable providers

rsafonseca:
Provides the ability for sharing a provider across namespaces.
Provides the ability for multiple alerts to share the same provider and post to different channels.

This avoids having to create a single provider for each channel created, reducing the number of necessary CRs when a provider can support multiple channels.

This avoids having to create a secret in every namespace that needs to use a provider to send notifications, which might not be desirable not only because of the sheer number of required objects, but also because the secret that the provider uses might not be something we want to share with owners of other namespaces.

Overall, this alleviates the usage of providers and alerts in large multi-tenant environments, as well as in smaller environments that might want to simplify their usage.

Addresses #125


rsafonseca:
@stefanprodan would you be able to take a look? 👀 
Thanks! :D

stefanprodan:
This needs an RFC https://github.com/fluxcd/flux2/tree/main/rfcs

rsafonseca:
Here you go @stefanprodan https://github.com/fluxcd/flux2/pull/4612 :)

ojsef39:
is there an update on this? :) thanks in advance!

PR #749, opened 2024-02-27
Remove `genclient:Namespaced` tag

gmolau:
This tag isn't used by controller-tools, only `nonNamespaced` is.

Context: https://cloud-native.slack.com/archives/CLAJ40HV3/p1708794732147909

Tested by running `make generate` and verifying that there is no diff.

### DIFF
--- a/api/v1/receiver_types.go
+++ b/api/v1/receiver_types.go
@@ -122,7 +122,6 @@ func (in *Receiver) GetInterval() time.Duration {
 }
 
 // +genclient
-// +genclient:Namespaced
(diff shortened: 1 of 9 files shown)


stefanprodan:
@gmolau can you please create an issue in flux2 for this task then link all the PRs to it. This way we don't lose track of them all.

gmolau:
@stefanprodan Sure, here https://github.com/fluxcd/flux2/issues/4634

(I'm waiting for all CIs to pass before opening the PRs now.)

stefanprodan (approved):

PR #757, opened 2024-03-07
Add the Provider `.spec.timeout` field to the API docs

stefanprodan:
The timeout field was missing from the API docs. 

Fix: https://github.com/fluxcd/flux2/issues/4653


stefanprodan:
Ah we don't pass the context to the http client, we need to fix this, good catch @darkowlzz!

stefanprodan:
@darkowlzz we actually do pass the context https://github.com/fluxcd/notification-controller/blob/e73db32382a1f60646d2540e8525eaa55ec66f5c/internal/notifier/client.go#L88-L90

stefanprodan:
> Then maybe we don't need to set the hard-coded timeout?

Agreed, this is leftover for when we didn't had timeout in the provider, but I high doubt setting it to zero changes anything.

PR #795, opened 2024-04-13
Alertmanager: Change timestamp label to .StartsAt

### LINKED ISSUE #793: Alerts sent to Alertmanager contain timestamps as labels, preventing alert grouping
# Description

notification-controller sets [`timestamp`][code] label on outgoing alerts, preventing Alertmanager from recognizing subsequent alerts with the same label set as a single alert (Alertmanager considers newly-posted alerts with an existing label set to be the same updated alert). This means Alertmanager will dispatch multiple alerts for what is essentially a single outage (e.g. failing `GitRepository` with a short reconciliation interval, e.g. `1m`, which is the default for `flux bootstrap`-generated `GitRepository`)

[code]: https://github.com/fluxcd/notification-controller/blob/0768aa25556bc456a2e33be1554bb96bb3a68c3b/internal/notifier/alertmanager.go#L78

This also requires us to add alert receivers with `send_resolved: false` separate from already-configured Prometheus receivers, as Flux alerts would be expiring at the default Alertmanager `resolve_timeout` of `5m`, with Flux posting new ones over time.

Alertmanager (as of v0.27.0) seems to lack the ability to drop incoming alert labels.

Multiple alerts in Alertmanager

# Proposal

While `group_by` combined with a long enough `group_interval` might make subsequent alerts less annoying, it still means we end up with multiple outgoing Alertmanager alerts in a single message, possibly grouped with the templates.

I propose removing this label altogether. Alertmanager already provides us with `.StartsAt` / `.EndsAt`, posting a new alert would just set the new `"startsAt"`. This would technically be a breaking change, as existing users might rely on this label in their alert templates.

I'm willing to work on a PR for this if needed.

# System details

Flux v2.2.3, notification-controller v1.2.4, Alertmanager v0.27.0

tdemin:
notification-controller posted all outgoing Alertmanager alerts with "timestamp" label, effectively preventing grouping alerts related to the same resource and forcing users to configure a separate alert receiver with `send_resolved: false`.

This changes it to instead set "startsAt", which was previously set (automatically by Alertmanager) to alert posting time. "endsAt" remains unset, as we have no way of figuring that out but the reconciliation interval of the resource that generated the alert, which can currently only be found out by making a Kubernetes API round-trip.

Note that this requires users to adapt alert templates that relied on .Labels.Timestamp.

Fixes #793.

### DIFF
--- a/docs/spec/v1beta3/providers.md
+++ b/docs/spec/v1beta3/providers.md
@@ -878,9 +878,10 @@ an [Event](events.md#event-structure) to the provided Prometheus Alertmanager
 [Address](#address).
 
 The Event will be formatted into a `firing` [Prometheus Alertmanager
-alert](https://prometheus.io/docs/alerting/latest/notifications/#alert),
-with the metadata added to the `labels` fields, and the `message` (and optional
-`.metadata.summary`) added as annotations.
+alert](https://prometheus.io/docs/alerting/latest/notifications/#alert), with
+the metadata added to the `labels` fields, and the `message` (and optional
+`.metadata.summary`) added as annotations. Event timestamp will be used to set
+alert start time (`.StartsAt`).
 
 In addition to the metadata from the Event, the following labels will be added:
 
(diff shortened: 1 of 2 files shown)


stefanprodan (commented):
Please update the docs here https://github.com/fluxcd/notification-controller/blob/main/docs/spec/v1beta3/providers.md#prometheus-alertmanager

tdemin:
@stefanprodan sorry for taking a while to update this. I've expanded the docs on this.

While we're at it, I also updated `fmt.Stringer` / `json.Marshaler` implementations of `AlertManagerTime` to use value receiver types instead of pointers, because this is what Go stdlib docs [tell us to do](https://pkg.go.dev/time#Time):

> Programs using times should typically store and pass them as values, not pointers. That is, time variables and struct fields should be of type time.Time, not *time.Time.

stefanprodan (approved):
LGTM

Thanks @tdemin 🏅

Issue #460, opened 2023-01-09
Opsgenie Alias for deduplication of alerts


hhollenstain:
We currently utilize Opsgenie for paging and found the integration with flux works pretty well. The main issue of contention is missing the alias for deduplication. Currently when an alert is triggered it will continuous fire/create new pages.  Ideally we can set an alias and fire once and let Opsgenie handle additional notification/triggers. 

https://github.com/fluxcd/notification-controller/blob/505345cb3ecc006ed71fae02d5988de7acd65ac9/internal/notifier/opsgenie.go#L68

[Opsgenie API docs](https://docs.opsgenie.com/docs/alert-api)

al-lac:
This would be nice to have.

Would also be great to have the alerts solve themself automatically once they are not happening anymore.

stefanprodan:
[event.Metadata](https://fluxcd.io/flux/components/notification/alerts/#event-metadata) gets injected in the payload sent to OpsGenie so you can add cluster name, region, etc. Can't this be used for deduplication?

PatrickZeier-SAG:
In JSM I see this in the created alert:
![image](https://github.com/fluxcd/notification-controller/assets/83637949/317877c0-df88-4c04-b53c-91ff1374a383)
(`summary`and `testField` were added by me in `spec.eventMetadata` of the Flux alert)

which according to the Jira API documentation matches this field:
![image](https://github.com/fluxcd/notification-controller/assets/83637949/e1c25e16-d405-4cce-921a-a117380d9190)

And `event.Metadata` seems to refer to `spec.eventMetadata` and is also part part of the payload:
```
payload := OpsgenieAlert{
		Message:     event.InvolvedObject.Kind + "/" + event.InvolvedObject.Name,
		Description: event.Message,
		Details:     event.Metadata,
```

With some Jira automation rules deduplication and title manipulation could work (need to check with some admin there on our side). Customization of the fields on Flux side would be a bit easier in my eyes, but see it as a feature request 😃
.
Many thanks @stefanprodan for the hint!

al-lac:
For Opsgenie i just set the alias to the description, works out most of the time, as long as the description does not contain a time string that is always different. 

What would be great however would be a message that could also close the alert. So like a "recovery" message.

stefanprodan:
Flux is stateless, there is no way to send recovery messages as notification-controller doesn't know it has send a previous error alert.

PatrickZeier-SAG:
[...]

PR #796, opened 2024-04-17
Add severity to opsgenie alerts

al-lac:
Add severity to opsgenie alerts to better be able to handle the opening and closing of alerts.

Related to #460

### DIFF
--- a/internal/notifier/opsgenie.go
+++ b/internal/notifier/opsgenie.go
@@ -65,10 +65,17 @@ func (s *Opsgenie) Post(ctx context.Context, event eventv1.Event) error {
 		return nil
 	}
 
+	var details = make(map[string]string)
+
+	if event.Metadata != nil {
+		details = event.Metadata
+	}
+	details["severity"] = event.Severity
+
 	payload := OpsgenieAlert{
 		Message:     event.InvolvedObject.Kind + "/" + event.InvolvedObject.Name,
 		Description: event.Message,
-		Details:     event.Metadata,
+		Details:     details,
 	}
 
 	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, func(req *retryablehttp.Request) {
--- a/internal/notifier/opsgenie_test.go
+++ b/internal/notifier/opsgenie_test.go
@@ -24,6 +24,7 @@ import (
 	"net/http/httptest"
 	"testing"
 
+	"github.com/fluxcd/pkg/apis/event/v1beta1"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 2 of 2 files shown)


internal/notifier/opsgenie.go
```diff
@@ -69,6 +70,7 @@ func (s *Opsgenie) Post(ctx context.Context, event eventv1.Event) error {
 		Message:     event.InvolvedObject.Kind + "/" + event.InvolvedObject.Name,
 		Description: event.Message,
 		Details:     event.Metadata,
+		Severity     event.Severity,
```
souleb, line 73:
```suggestion
		Severity:     event.Severity,
```

PatrickZeier-SAG: I am not sure if this will work. For OpsGenie/JSM to use this field, it must go under the `Details` section (means `extraProperties` in Flux). Like `revision`. But I am not so familiar with golang, so don't know how to implement.

al-lac: Yeah i was wondering how to test this. I think it will just be added as a separate dynamic property, but i may be wrong.

PatrickZeier-SAG: Just check the [JIRA API docs](https://developer.atlassian.com/cloud/jira/service-desk-ops/rest/v1/api-group-integration-events/#api-jsm-ops-integration-v2-alerts-post) for the available body fields. OG/Jira works with `priority`.

stefanprodan:
Please run `make test` before committing, the code is not properly formatted.

internal/notifier/opsgenie.go
PatrickZeier-SAG: > Yeah i was wondering how to test this. I think it will just be added as a separate dynamic property, but i may be wrong.

As the rest of the fields here in Flux matches the ones in the Jira API, I think this is a 1:1 relation.

And I think I missed the point that anyway the whole `event.Metadata` (means `spec.extraProperties` section) is inserted. So the main issue is IMHO: In this Flux opsgenie module section, it can only put it as separate parameter to the body. JSM can only access the keys under `Details` for usage in a rule. Maybe there is some other place where `event.Metadata` is created and there the severity can be added.

al-lac:
@stefanprodan yeah i think i need to rework this. Give me some time to come up with something else.

al-lac:
Added the severity now to the details payload, this should hopefully be the easiest solution.

internal/notifier/opsgenie.go
```diff
@@ -71,6 +71,9 @@ func (s *Opsgenie) Post(ctx context.Context, event eventv1.Event) error {
 		Details:     event.Metadata,
 	}
 
+	// Add severity to the details map.
+	payload.Details["severity"] = event.Severity
```
souleb, line 75: we have to watch for a `nil`map.

I would do something like

```go
	var details = make(map[string]string)
	if event.Metadata != nil {
		details = event.Metadata
	}
	details["severity"] = event.Severity

	payload := OpsgenieAlert{
		Message:     event.InvolvedObject.Kind + "/" + event.InvolvedObject.Name,
		Description: event.Message,
                Details: details,
	}
```

I would also add a test for this.

al-lac: Added it, not sure how to properly write a test for this though.

souleb:
@al-lac I have added the test. Can you test that this works in your cluster?

al-lac:
@souleb works in my cluster and shows the severity in opsgenie now 🎉 
CleanShot 2024-04-18 at 15 57 02@2x

souleb (approved):
LGTM! Thanks @al-lac

souleb:
@al-lac can you squash your commits please?

souleb:
Please also update the commit message.

al-lac:
@souleb i hope it is fine like this now.

PR #814, opened 2024-05-02
Update Helm APIs to `source.toolkit.fluxcd.io/v1` (GA)

stefanprodan:
xref: https://github.com/fluxcd/source-controller/pull/1428


swade1987 (approved):

souleb (approved):
LGTM!

Issue #828, opened 2024-05-15
Metrics has a potential cardinality issue


Kuzbekov:
By nature, our webhooks are usually open for external connections, especially when the code is on cloud-based repositories. If somebody tries to scan the hostname for webhook directed to the webhook receiver for vulnerabilities and attempts to gain access to multiple paths, it will generate a lot of metrics records and lead to cardinality explosion on the Prometheus server because the label "handle" stores the path from each new request due to dynamic handler naming.
It can be mitigated by making ingress with static path, but it's not mentioned in documentation.
Relevant information is  [here](https://github.com/slok/go-http-metrics/?tab=readme-ov-file#custom-handler-id)

souleb:
We would need a new flag for the controller and changes in https://github.com/fluxcd/notification-controller/blob/580497beeb8bee4cee99163bb63fba679cd2d735/internal/server/event_server.go#L86 and https://github.com/fluxcd/notification-controller/blob/580497beeb8bee4cee99163bb63fba679cd2d735/internal/server/receiver_server.go#L53

Are you willing to contribute the change?

Kuzbekov:
I would like to contribute. Do you think it would be better to implement a flag with a default value that ensures safety? This change would necessitate notifying users due to its potential impact. Currently, someone could theoretically use precise webhook values in the 'handle' label within alerts or metrics. Or  we should keep current behavior?

souleb:
we have to keep the default behavior. Will should add recommendation in the documentation.

stefanprodan:
Fixed in Flux 2.4

PR #841, opened 2024-05-29
New flag to disable detailed metrics for path

Kuzbekov:
Flag detailed-metrics added to provide a way to disable exposing all accessed paths to the metrics and  prevent potential metrics cardinality explosion

Issue ref: https://github.com/fluxcd/notification-controller/issues/828

### DIFF
--- a/internal/server/event_server.go
+++ b/internal/server/event_server.go
@@ -52,17 +52,19 @@ type EventServer struct {
 	logger               logr.Logger
 	kubeClient           client.Client
 	noCrossNamespaceRefs bool
+	detailedMetrics      bool
 	kuberecorder.EventRecorder
 }
 
 // NewEventServer returns an HTTP server that handles events
-func NewEventServer(port string, logger logr.Logger, kubeClient client.Client, eventRecorder kuberecorder.EventRecorder, noCrossNamespaceRefs bool) *EventServer {
+func NewEventServer(port string, logger logr.Logger, kubeClient client.Client, eventRecorder kuberecorder.EventRecorder, noCrossNamespaceRefs bool, detailedMetrics bool) *EventServer {
 	return &EventServer{
 		port:                 port,
 		logger:               logger.WithName("event-server"),
 		kubeClient:           kubeClient,
 		EventRecorder:        eventRecorder,
 		noCrossNamespaceRefs: noCrossNamespaceRefs,
+		detailedMetrics:      detailedMetrics,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -97,6 +98,7 @@ func main() {
 	flag.BoolVar(&watchAllNamespaces, "watch-all-namespaces", true,
 		"Watch for custom resources in all namespaces, if set to false it will only watch the runtime namespace.")
 	flag.DurationVar(&rateLimitInterval, "rate-limit-interval", 5*time.Minute, "Interval in which rate limit has effect.")
+	flag.BoolVar(&detailedMetrics, "detailed-metrics", true, "Count metrics for every requested path")
```
stefanprodan, line 101: I would name the flag `--export-http-path-metrics` and have it set to `false` by default. The description could be: `When enabled, the requests full path is included in the HTTP server metrics (risk as high cardinality).`

souleb: it is the other way around.

maybe something like:

`--set-metrics-handlerID` and have it set to false by default. The description: When enabled, the requests full path is included in the HTTP server metrics as handler label (this will keep cardinality low because all urls path will have same handler label on the metrics).

Like this we keep the actual behavior by default.

internal/server/event_server.go
```diff
@@ -82,8 +84,13 @@ func (s *EventServer) ListenAndServe(stopCh <-chan struct{}, mdlw middleware.Mid
 		handler = middleware(handler)
 	}
 	mux := http.NewServeMux()
-	mux.Handle("/", handler)
-	h := std.Handler("", mdlw, mux)
+	path := "/"
+	mux.Handle(path, handler)
+	middlewarePath := ""
```
souleb, line 89:
```suggestion
	handlerID := ""
```

main.go
stefanprodan: But I don't think we should keep the default behaviour, given that people expose the receiver on the internet, best is to ship with a secure default.

Kuzbekov: I agree with @stefanprodan, but the only concern is if somebody have already build alerts system based on path labels that they have in their systems. Probability is low, but still this change would silently break metrics and affect monitoring for those who rely on this label. Probably we could keep current behaviour and switch to more secure on next major release with proper notification in release notes  ?

stefanprodan: This PR will not go into a patch release anyway. It will be part of Flux v2.4 and we'll document the flag and the behaviour change accordingly.

stefanprodan:
@Kuzbekov can you please rebase and force push, we're getting close to Flux 2.4 release and this PR can now be merged.

Kuzbekov:
> @Kuzbekov can you please rebase and force push, we're getting close to Flux 2.4 release and this PR can now be merged.

done

stefanprodan (approved):
LGTM

Thanks @Kuzbekov 🏅

PR #894, opened 2024-08-01
Fix telegram test flake

### LINKED ISSUE #867: Telegram test is flaky
https://github.com/fluxcd/notification-controller/blob/39f345231ffbb699d41aa129fe9f28d56a207c0c/internal/notifier/telegram_test.go#L46 fails sometimes because the metadata string is assembled from a map where the iteration order is non-deterministic.

makkes:
The test would sometimes fail because the metadata lines are built from a map and map iteration order in Go is non-deterministic.  Therefore, the lines may be ordered differently between different test runs.

Now, the test sorts the metadata lines of the telegram message so they are always the same and only then verifies the expected output.

closes #867

### DIFF
--- a/internal/notifier/telegram_test.go
+++ b/internal/notifier/telegram_test.go
@@ -22,6 +22,8 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"slices"
+	"strings"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


internal/notifier/telegram_test.go
```diff
@@ -52,3 +65,10 @@ func TestTelegram_Post(t *testing.T) {
 	err = telegram.Post(context.TODO(), ev)
 	require.NoError(t, err)
 }
+
+/*
```
matheuscscp, line 69: Is this comment left here on purpose?

makkes: Oops, no. I removed it now.

matheuscscp (approved):
LGTM!

stefanprodan (approved):

Issue #491, opened 2023-03-09
Filtering for incoming webhooks


tun0:
A more generic approach similar to the support for filtering on Events for GitHub, etc.

For example GCR/GAR sends a pretty simple, but also useful payload: 
```json
{
  "action":"INSERT",
  "digest":"us-east1-docker.pkg.dev/my-project/my-repo/hello-world@sha256:6ec128e26cd5...",
  "tag":"us-east1-docker.pkg.dev/my-project/my-repo/hello-world:1.1"
}
```
Ref: https://cloud.google.com/artifact-registry/docs/configure-notifications#examples

It would be nice to be able to map the digest to a ImageRepository to notify (either automatically or explicitly configured). This could potentially reduce the load within and caused by Flux drastically.

The current alternative (to reduce the load) is to configure separate incoming webhooks on both GCR/GAR's and Flux' side. Which is rather cumbersome.

makkes:
I'm not sure I'm following what you're requesting. Would you mind please describing the envisioned API changes and/or workflow you're aiming at?

tun0:
Currently we have a single receiver to consume notifications from our GCR/GAR. In this receiver `spec.resources` lists 34 `ImageRepository` references. Whenever a notification arrives from GCR or GAR, all 34 image repositories get annotated and scanned for new images. What I'm proposing is instead of triggering a scan for all 34, only trigger those image repositories that match against the `digest`part of the notification payload  (without the `@...` part).

lucioveloso:
I will go even deep. The primary concern isn't just that all images are annotated for reconciliation. The core issue is the reconciliation process itself. I have thousands of images, and with each new webhook received, there's a scan of all image tags in Harbor. This results in significant resource consumption in Harbor instance.

stefanprodan:
Each provider has a unique payload schema, we'll need to get all those schemas as dependencies (does al of them have Go SDKs?). After that we'll need to write a custom parser for each one, to be able to extract some string that may or may not match the Flux resource name listed under `.spec.resources`.  The Receiver API is common to all current and future Flux resource types which makes this issue about container images so much larger is scope.

bigkevmcd:
Could something like CEL be used to process the incoming request body (assuming JSON).

Taking the example above...

```json
{
  "action":"INSERT",
  "digest":"us-east1-docker.pkg.dev/my-project/my-repo/hello-world@sha256:6ec128e26cd5...",
[...]

PR #928, opened 2024-09-17
Initial version of CEL resource expressions.

bigkevmcd:
This is a very rough cut (needs additional error checking).

But the basic implementation works.

This is the code I referenced here https://github.com/fluxcd/notification-controller/issues/491#issuecomment-2305313021

I will harden it up, checking that it's receiving a `json` body and improve the error handling, this was a quick bit of code to see if it was possible.


api/v1/receiver_types.go
```diff
@@ -63,10 +63,24 @@ type ReceiverSpec struct {
 	// +optional
 	Events []string `json:"events,omitempty"`
 
+	// TODO: Validate one or other (or both?)
+
```
bigkevmcd, line 67: This will need to be addressed, because you could have `.spec.resourceExpressions` or `.spec.resources` (or both).

matheuscscp (commented):
Thanks for the work here @bigkevmcd, but I think we need to create an RFC or at least an issue to discuss this feature.

internal/server/receiver_handlers.go
```diff
@@ -544,3 +565,135 @@ func getGroupVersion(s string) (string, string) {
...
+		// TODO Log out other types?
+	}
+
+	return nil, nil
+}
+
+func parseList(l traits.Lister) ([]apiv1.CrossNamespaceObjectReference, error) {
+	var resources []apiv1.CrossNamespaceObjectReference
+	it := l.Iterator()
+	for it.HasNext().Value() == true {
```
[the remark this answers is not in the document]
matheuscscp: I made this suggestion thinking the type of `it.HasNext().Value()` was `bool`, but it's actually `any`, so looks like you do need `== true` here. I didn't know you could compare `any` with `bool` :thinking:

matheuscscp:
> Thanks for the work here @bigkevmcd, but I think we need to create an RFC or at least an issue to discuss this feature.

Looking again at the original issue, maybe we could use that issue to discuss this feature, it looks like in this PR you're specifically trying to address this part:

> It would be nice to be able to map the digest to a ImageRepository to notify (either automatically or explicitly configured). This could potentially reduce the load within and caused by Flux drastically.

internal/server/receiver_handlers.go
```diff
@@ -544,3 +565,135 @@ func getGroupVersion(s string) (string, string) {
...
+func (s *ReceiverServer) evaluateResourceExpressions(r *http.Request, receiver apiv1.Receiver) ([]apiv1.CrossNamespaceObjectReference, error) {
+	if len(receiver.Spec.ResourceExpressions) == 0 {
+		return nil, nil
+	}
+
+	body := map[string]any{}
+	// Only decodes the body for the expression if the body is JSON.
+	// Technically you could generate several resources without any body.
+	if isJSONContent(r) {
+		if err := json.NewDecoder(r.Body).Decode(&body); err != nil {
```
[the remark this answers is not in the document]
bigkevmcd: I missed this in the code above, I'll fix this.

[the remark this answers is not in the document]
matheuscscp: Another option is simply reading the body into a `bytes.Buffer` and passing it to all the parts of the code that need to consume it.

matheuscscp:
To fix the e2e GitHub Action failing you can run `make api-docs manifests generate` locally to generate the missing change for `docs/api/v1/notification.md`.

matheuscscp:
Now there's a diff on `config/crd/bases/notification.toolkit.fluxcd.io_receivers.yaml` breaking the e2e GitHub Action. When I run `make manifests` (or my usual `make api-docs manifests generate`), I see the missing diff on this file locally. Can you please run it and see if it generates this missing diff? Thanks!

bigkevmcd:
> > Thanks for the work here @bigkevmcd, but I think we need to create an RFC or at least an issue to discuss this feature.
That's ok, it was a quick branch to explore what's possible, there's no compelling need to land this, and, it's taking on support for CEL, which is not "free".

Worst case, it gets closed as "won't fix" 😄 and informs a different solution, fine by me.

> > It would be nice to be able to map the digest to a ImageRepository to notify (either automatically or explicitly configured). This could potentially reduce the load within and caused by Flux drastically.

Yes, this is basically a way to dynamically generate the set of resources to notify in a receiver.

The new field `.spec.resourceExpressions` is basically a set of CEL expressions that _can_ return a the resources that should be notified, it could be simplified to a single `resourceExpression` but from experience with CEL elsewhere, this tends to lead to complex expressions.

matheuscscp:
This solution sounds very elegant to me @bigkevmcd, I'd love to see it in. It would be nice if you could attend some of our dev meetings to discuss this feature, they are public and can be found here: https://fluxcd.io/community/#meetings

(The `CNCF Flux Project Meeting` ones.)

bigkevmcd:
I'll see what I can do to be there, time permitting.

You'll see that the specific request in here https://github.com/fluxcd/notification-controller/issues/491#issue-1617811745 is tested in the code.

api/v1/receiver_types.go
```diff
@@ -63,10 +63,22 @@ type ReceiverSpec struct {
...
 
+	// ResourceExpressions is a list of CEL expressions that will be parsed to
+	// determine resources to be notified about changes.
+	// The expressions must evaluate to CEL values that contain the keys "name",
+	// "kind", "apiVersion" and optionally "namespace".
+	//  These values will be parsed to CrossNamespaceObjectReferences.
+	// e.g. {"name": "test-resource-1", "kind": "Receiver", "apiVersion":
+	//       "notification.toolkit.fluxcd.io/v1"}.
+	// +optional
+	ResourceExpressions []string `json:"resourceExpressions,omitempty"`
```
stefanprodan, line 80: I think this should be named `ResourceFilters` and act on the given list of `Resources`.

bigkevmcd: @stefanprodan The last test is borked because of a bug in controller-runtime which I've fixed, but is yet to be released

If I understand what you're saying here, you would have `Resources` and the user would declare all the possible resources, and the `ResourceFilters` would filter these to only a subset based on the request?

This wouldn't be dynamic tho', it's not clear how it would solve the problem outlined in #491?

matheuscscp: > If I understand what you're saying here, you would have Resources and the user would declare all the possible resources, and the ResourceFilters would filter these to only a subset based on the request?

Precisely, that's exactly what we discussed in the dev meeting today!

matheuscscp: > This wouldn't be dynamic tho', it's not clear how it would solve the problem outlined in https://github.com/fluxcd/notification-controller/issues/491?

Our reasoning is security, this style forces a more secure stance, the user always have to the declare which resources can be reconciled by the Receiver. They may end up with a long list of resources but at least it will always be clear from the Receiver object which Flux objects can be reconciled from that Receiver.

PR #931, opened 2024-09-25
Update fluxcd/pkg deps and k8s deps to 1.31.1

darkowlzz:


matheuscscp (approved):

stefanprodan (changes requested):
The api package needs to be updated too, also the controller go.mod should be on the same version on SC go.mod which is `go 1.22.5`.

stefanprodan (approved):

PR #953, opened 2024-10-14
Clarify gitlab notification provider usage

swoehrl-mw:
The currrent gitlab v4 API does no longer accept a project name for setting git commit status, a project ID must be used.

This PR adds an explanation and example for that to the provider docs and also clarifies the scope for the access token.


matheuscscp (approved):

matheuscscp:
Thanks @swoehrl-mw !

PR #997, opened 2024-12-08
feat: Remove deprecated object metrics from controllers fluxcd/flux2#5083

hasithsen:
Suspend, readiness, and duration metrics recording removed from Receiver reconciler.


stefanprodan:
@hasithsen please squash

PR #1010, opened 2024-12-19
fix: adding of duplicate commit statuses in gitlab

### LINKED ISSUE #1009: GitLab Provider adds commit status for every reconciliation
The GitLab Provider continuously adds commit statuses for every reconciliation.

This happens only for projects with more than 20 jobs. The GitLab Provider code already has some logic to prevent adding a commit status if one already exists. Still, it only considers the first 20 commit statuses because it is a paginated request that defaults to 20.

The impact of this bug is quite heavy. Thousands of commit statuses considered "external jobs" in the GitLab UI have piled up, with every new job triggering some GitLab background processes, nearly bringing down our GitLab instance.

mo-rieger:
The get commit-status request is paginated and only returns the first 20 commit statuses by default. This change fetches the commit-status by name to avoid multiple requests.

RESOLVES #1009

### DIFF
--- a/internal/notifier/gitlab.go
+++ b/internal/notifier/gitlab.go
@@ -99,7 +99,9 @@ func (g *GitLab) Post(ctx context.Context, event eventv1.Event) error {
 		Description: desc,
 	}
 
-	getOpt := &gitlab.GetCommitStatusesOptions{}
+	getOpt := &gitlab.GetCommitStatusesOptions{
+		Name: &status.Name,
+	}
 	statuses, _, err := g.Client.Commits.GetCommitStatuses(g.Id, rev, getOpt, gitlab.WithContext(ctx))
 	if err != nil {
 		return fmt.Errorf("unable to list commit status: %s", err)


matheuscscp:
@mo-rieger Can you please build an image, test this e2e and post some evidence that it works? Then I think we can approve and merge the PR. Thanks!

mo-rieger:
Hi @matheuscscp, I built the [image](https://hub.docker.com/repository/docker/mowglee/notification-controller/general) and replaced the notification-controller with the new built.
The patch fixes the bug. The GitLab commit statuses get updated but the GitLab Jobs did not increase over time like it was the case before.
image
Do you need anything else?

stefanprodan:
Thanks @mo-rieger for testing this end-to-end 🥇 I expected for GitLab to error out when the commit status doesn't exists, but I guess they just return an empty list so the first time an event is issued, NC will create the status.

matheuscscp (approved):
Thanks very much for this contribution, @mo-rieger!

mo-rieger:
> I expected for GitLab to error out when the commit status doesn't exists, but I guess they just return an empty list so the first time an event is issued, NC will create the status.

Exactly, it's always 200 with an empty list if no commit status exist.

Thanks for merging!

PR #1014, opened 2024-12-24
[RFC-0008] Custom Event Metadata from Annotations

matheuscscp:
Part of: https://github.com/fluxcd/flux2/pull/4809

The most significant change in this PR for implementing RFC 0008 is moving the group prefix removal from the `cleanupMetadata()` function deeper into the request handling logic, such that the precedence order defined in the RFC can be concisely implemented in a single place. This is because different Alert objects matching an event may append different Alert-level metadata to the event, and hence for the metadata combination, precedence order and key conflict detection to be implemented in a single place, moving the group prefix removal deeper was necessary. This logic is now implemented by the function `combineEventMetadata()`, which is an evolution of the renamed `enhanceEventWithAlertMetadata()`. As a result, a few other functions manipulating metadata keys now have to consider the involved object group prefix when doing so.

Tasks:
* [x] Use constant `Group` from `github.com/fluxcd/pkg/apis/event` after https://github.com/fluxcd/pkg/pull/848 is merged and the version can be bumped.


docs/spec/v1beta3/alerts.md
```diff
@@ -168,9 +177,68 @@ spec:
   inclusionList:
     - ".*succeeded.*"
   eventMetadata:
-    app.kubernetes.io/env: "production"
-    app.kubernetes.io/cluster: "my-cluster"
-    app.kubernetes.io/region: "us-east-1"
+    env: production
+    cluster: my-cluster
+    region: us-east-1
```
matheuscscp, line 182: I'm removing the `app.kubernetes.io/` prefix here since it has absolutely no effect and may only cause confusion in the presence of other more important prefixes the user may need to understand, e.g. `event.toolkit.fluxcd.io/`.

internal/server/event_handlers.go
```diff
@@ -418,30 +419,90 @@ func (s *EventServer) eventMatchesAlertSource(ctx context.Context, event *eventv
...
+		metadata[k] = v
+		metadataSources[k] = append(metadataSources[k], sourceAlertEventMetadata)
 	}
 
+	// 3) Alert .spec.summary with the key "summary".
 	if alert.Spec.Summary != "" {
-		meta["summary"] = alert.Spec.Summary
+		metadata[summaryKey] = alert.Spec.Summary
+		metadataSources[summaryKey] = append(metadataSources[summaryKey], sourceAlertSummary)
+		l.Info("warning: specifying alert summary cert via '.spec.summary' is deprecated, please use '.spec.eventMetadata.summary' instead")
```
stefanprodan, line 471:
```suggestion
		l.Info("warning: specifying an alert summary with '.spec.summary' is deprecated, use '.spec.eventMetadata.summary' instead")
```

matheuscscp: fixed, thanks!

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1018, opened 2025-01-20
Add support for MetaOriginRevisionKey from the Event API

matheuscscp:
Part of: https://github.com/fluxcd/flux2/issues/5148

### DIFF
--- a/internal/server/event_server.go
+++ b/internal/server/event_server.go
@@ -233,6 +233,12 @@ func eventKeyFunc(r *http.Request) (string, error) {
 
 	objectGroup := event.InvolvedObject.GetObjectKind().GroupVersionKind().Group
 
+	originRevisionKey := fmt.Sprintf("%s/%s", objectGroup, eventv1.MetaOriginRevisionKey)
+	originRevision, ok := event.Metadata[originRevisionKey]
+	if ok {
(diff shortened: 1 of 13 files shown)


internal/server/event_server.go
```diff
@@ -233,9 +233,11 @@ func eventKeyFunc(r *http.Request) (string, error) {
 
 	objectGroup := event.InvolvedObject.GetObjectKind().GroupVersionKind().Group
 
+	originRevisionKey := fmt.Sprintf("%s/%s", objectGroup, eventv1.MetaOriginRevisionKey)
 	revisionKey := fmt.Sprintf("%s/%s", objectGroup, eventv1.MetaRevisionKey)
-	revision, ok := event.Metadata[revisionKey]
-	if ok {
+	if originRevision, ok := event.Metadata[originRevisionKey]; ok {
+		comps = append(comps, "revision="+originRevision)
+	} else if revision, ok := event.Metadata[revisionKey]; ok {
```
stefanprodan, line 240: We should add both to the rate limiter key, the origin revision may be the same while the OCI digest may change.

matheuscscp: fixed, thanks

stefanprodan (approved):
LGTM

Thanks @matheuscscp 🏅

Issue #1020, opened 2025-01-24
Adding headers to requests for Alertmanager provider.


d4rkfella:
Hello, i've been trying to create alertmanager provider using oauth as an authentication method. What i did was create a docker image that runs a script to obtain an access token then create a secret with the obtained token like this . 
```
#!/bin/bash

TOKEN=$(curl -X POST "$KEYCLOAK_URL" \
  -d "client_id=$CLIENT_ID" \
  -d "client_secret=$CLIENT_SECRET" \
  -d "grant_type=client_credentials" \
  -d "scope=$SCOPE" \
  -H "Content-Type: application/x-www-form-urlencoded" | jq -r .access_token)

echo -e "apiVersion: v1
kind: Secret
metadata:
  name: oidc-jwt
stringData:
  headers: |
    Authorization: Bearer $TOKEN" | kubectl apply -f -
```
I've manually tested the token with curl and i can authenticate no problem, but when the notification controller tries to send an alert it not adding the header i have specified in the referenced secret. Is it currently possible to add headers to the requests for an Alertmanager provider ?

matheuscscp:
The Provider API already supports the specification of a token inside a secret:

https://fluxcd.io/flux/components/notification/providers/#token-example

But the `alertmanager` provider does not use it. We could modify the constructor of this provider to accept a token like `opsgenie`:

https://github.com/fluxcd/notification-controller/blob/fa7d9f260c00f1325df0ee6dfc66a924ee991753/internal/notifier/opsgenie.go#L43-L57

And, if the token is not empty, use it more or less like this:

https://github.com/fluxcd/notification-controller/blob/fa7d9f260c00f1325df0ee6dfc66a924ee991753/internal/notifier/opsgenie.go#L81-L83

It's an easy implementation. @d4rkfella are you willing to contribute this feature?

d4rkfella:
I could give it a shot . Looks easy enough for me :P

d4rkfella:
Opened a pull request , let me know if i did it correctly.

PR #1021, opened 2025-01-25
Add support for Bearer Token authentication to Provider `alertmanager`

d4rkfella:
I'm not a programmer, but after examining the other providers, i think this should be correct. Please let me know :)

Closes: #1020

### DIFF
--- a/docs/spec/v1beta3/providers.md
+++ b/docs/spec/v1beta3/providers.md
@@ -919,10 +919,13 @@ and [TLS certificates](#tls-certificates).
 
 ###### Prometheus Alertmanager example
 
-To configure a Provider for Prometheus Alertmanager, create a Secret with [the
-`address`](#address-example) set to the Prometheus Alertmanager [HTTP API
+To configure a Provider for Prometheus Alertmanager, authentication can be done using either Basic Authentication or a Bearer Token. 
+Both methods are supported, but using authentication is optional based on your setup.
+
+Basic Authentication: 
+Create a Secret with [the `address`](#address-example) set to the Prometheus Alertmanager [HTTP API
 URL](https://prometheus.io/docs/alerting/latest/https/#http-traffic)
-including Basic Auth credentials, and a `alertmanager` Provider with a [Secret
+including Basic Auth credentials, and an `alertmanager` Provider with a [Secret
 reference](#secret-reference).
 
 ```yaml
@@ -943,7 +946,33 @@ metadata:
   name: alertmanager-address
   namespace: default
 stringData:
-    address: https://username:password@<alertmanager-url>/api/v2/alerts/"
+    address: https://<username>:<password>@<alertmanager-hostport>/api/v2/alerts/
+```
+Bearer Token Authentication:
+Create a Secret with [the `token`](#token-example), and an `alertmanager` Provider with a [Secret
+reference](#secret-reference) and the Prometheus Alertmanager [HTTP API
+URL](https://prometheus.io/docs/alerting/latest/https/#http-traffic) set directly in the `.spec.address` field.
+
+```yaml
+---
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Provider
+metadata:
+  name: alertmanager
+  namespace: default
+spec:
+  type: alertmanager
+  address: https://<alertmanager-hostport>/api/v2/alerts/
+  secretRef:
+    name: alertmanager-token
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: alertmanager-token
+  namespace: default
+stringData:
+    token: <token>
 ```
 
 ##### Webex
--- a/internal/notifier/alertmanager.go
+++ b/internal/notifier/alertmanager.go
@@ -24,6 +24,7 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
 
@@ -34,6 +35,7 @@ type Alertmanager struct {
 	URL      string
 	ProxyURL string
 	CertPool *x509.CertPool
+	Token    string
 }
 
 type AlertManagerAlert struct {
@@ -72,7 +74,7 @@ func (a *AlertManagerTime) UnmarshalJSON(jsonRepr []byte) error {
 	return nil
 }
 
-func NewAlertmanager(hookURL string, proxyURL string, certPool *x509.CertPool) (*Alertmanager, error) {
+func NewAlertmanager(hookURL string, proxyURL string, certPool *x509.CertPool, token string) (*Alertmanager, error) {
 	_, err := url.ParseRequestURI(hookURL)
 	if err != nil {
 		return nil, fmt.Errorf("invalid Alertmanager URL %s: '%w'", hookURL, err)
@@ -82,6 +84,7 @@ func NewAlertmanager(hookURL string, proxyURL string, certPool *x509.CertPool) (
 		URL:      hookURL,
 		ProxyURL: proxyURL,
 		CertPool: certPool,
+		Token:    token,
 	}, nil
 }
 
@@ -134,8 +137,13 @@ func (s *Alertmanager) Post(ctx context.Context, event eventv1.Event) error {
 		},
 	}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload)
-
+	var opts []requestOptFunc
+	if s.Token != "" {
+		opts = append(opts, func(request *retryablehttp.Request) {
+			request.Header.Add("Authorization", "Bearer "+s.Token)
+		})
+	}
+	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, opts...)
 	if err != nil {
 		return fmt.Errorf("postMessage failed: %w", err)
 	}
--- a/internal/notifier/alertmanager_fuzz_test.go
+++ b/internal/notifier/alertmanager_fuzz_test.go
@@ -46,7 +46,7 @@ func Fuzz_AlertManager(f *testing.F) {
 		var cert x509.CertPool
 		_ = fuzz.NewConsumer(seed).GenerateStruct(&cert)
 
-		alertmanager, err := NewAlertmanager(fmt.Sprintf("%s/%s", ts.URL, urlSuffix), "", &cert)
+		alertmanager, err := NewAlertmanager(fmt.Sprintf("%s/%s", ts.URL, urlSuffix), "", &cert, "")
 		if err != nil {
 			return
 		}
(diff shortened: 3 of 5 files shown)


internal/notifier/alertmanager.go
```diff
@@ -82,6 +84,7 @@ func NewAlertmanager(hookURL string, proxyURL string, certPool *x509.CertPool) (
 		URL:      hookURL,
 		ProxyURL: proxyURL,
 		CertPool: certPool,
+		Token: token,
```
matheuscscp, line 87:
```suggestion
		Token:    token,
```

We use [`go fmt`](https://go.dev/blog/gofmt) to format files, struct fields are always aligned using whitespaces.

internal/notifier/alertmanager.go
```diff
@@ -28,12 +28,14 @@ import (
 	"golang.org/x/text/language"
 
 	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/hashicorp/go-retryablehttp"
```
matheuscscp, line 31: Please put this import on the block right above. This is the order of our import blocks:

1. Built-in Go packages
2. Third-party packages
3. Flux packages

`"github.com/hashicorp/go-retryablehttp"` is a third-party.

Also, running `go fmt` would put it right on top of the other two third-party packages (`"golang.org/x/text/cases"` and `"golang.org/x/text/language"`).

internal/notifier/alertmanager.go
```diff
@@ -134,8 +137,11 @@ func (s *Alertmanager) Post(ctx context.Context, event eventv1.Event) error {
 		},
 	}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload)
-
+	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, func(request *retryablehttp.Request) {
+		if s.Token != "" {
+			request.Header.Add("Authorization", "Bearer "+s.Token)
+		}
+	})
```
matheuscscp, line 144:
```suggestion
	var opts []requestOptFunc
	if s.Token != "" {
		opts = append(opts, func(request *retryablehttp.Request) {
			request.Header.Add("Authorization", "Bearer "+s.Token)
		})
	}
	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, opts...)
```

Your original code would work correctly, but I believe putting the if outside the option makes more sense. In this suggestion, an option would only be passed if the token is not empty.

matheuscscp (commented):
Thanks for this contribution, @d4rkfella! Left a few comments.

We also need to update the docs to mention this feature. Please update this file:

https://github.com/fluxcd/notification-controller/blob/fa7d9f260c00f1325df0ee6dfc66a924ee991753/docs/spec/v1beta3/providers.md?plain=1#L922-L947

Please mention that the authentication for this provider can now be done through two options: basic auth (the one already documented), and bearer token. You can duplicate the example of basic auth and adapt it for the token. In this case, instead of setting the `address` field in the secret, this field would be set directly in the `Provider` object, `.spec.address`, and the secret would container the `token` field instead.

d4rkfella:
Ok i made the changes you suggested, took me a while to figure out how to fix the Signed off by message hah .

docs/spec/v1beta3/providers.md
```diff
@@ -945,6 +948,32 @@ metadata:
 stringData:
     address: https://username:password@<alertmanager-url>/api/v2/alerts/"
 ```
+Bearer Token Authentication:
+Create a Secret with [the `token`](#token-example), and a `alertmanager` Provider with a [Secret
```
matheuscscp, line 952:
```suggestion
Create a Secret with [the `token`](#token-example), and an `alertmanager` Provider with a [Secret
```

If the word following an indefinite article starts with the sound of a vowel, like "alert", the article is `an`. :grin: 

It's not part of your change but please fix this on the basic auth docs right above as well :pray: Much appreciated :grin:

docs/spec/v1beta3/providers.md
```diff
@@ -945,6 +948,32 @@ metadata:
...
+reference](#secret-reference) and Prometheus Alertmanager [HTTP API
+URL](https://prometheus.io/docs/alerting/latest/https/#http-traffic) set directly in the spec.address field.
```
matheuscscp, line 954:
```suggestion
URL](https://prometheus.io/docs/alerting/latest/https/#http-traffic) set directly in the `.spec.address` field.
```

docs/spec/v1beta3/providers.md
```diff
@@ -945,6 +948,32 @@ metadata:
 stringData:
     address: https://username:password@<alertmanager-url>/api/v2/alerts/"
 ```
+Bearer Token Authentication:
+Create a Secret with [the `token`](#token-example), and a `alertmanager` Provider with a [Secret
+reference](#secret-reference) and Prometheus Alertmanager [HTTP API
```
matheuscscp, line 953:
```suggestion
reference](#secret-reference) and the Prometheus Alertmanager [HTTP API
```

docs/spec/v1beta3/providers.md
```diff
@@ -945,6 +948,32 @@ metadata:
...
+```yaml
+---
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Provider
+metadata:
+  name: alertmanager
+  namespace: default
+spec:
+  type: alertmanager
+  address: <alertmanager-url>/api/v2/alerts/
```
matheuscscp, line 965:
```suggestion
  address: https://<alertmanager-hostport>/api/v2/alerts/
```

Would you please kindly also fix this in the basic auth example? :pray: Much appreciated!

`https://username:password@<alertmanager-url>/api/v2/alerts/"`

Becomes:

`https://<username>:<password>@<alertmanager-hostport>/api/v2/alerts/`

matheuscscp (commented):
Thanks for working through my comments :pray: 

There's a final batch, after those I think we won't need any more changes :grin: 

Before we merge this, I'll ask you to please kindly test this change in your environment and post here evidence that it works, e.g. logs, screenshots, etc.

Here are the steps to test this:

1. If you haven't done this already, please [generate a GitHub Personal Access Token (Classic)](https://github.com/settings/tokens/new) with the `write:packages` permission in order to push images to the GitHub Container Registry (GHCR). Login locally in Docker using this token: `docker login -u d4rkfella ghcr.io --password-stdin` and paste the token in the prompt.
2. Edit the project `Makefile` in the root directory changing the image like this:

```
IMG ?= fluxcd/notification-controller:latest
```

Becomes:

```
IMG ?= ghcr.io/d4rkfella/fluxcd/notification-controller:am-bearer
```

3. Build an image using the following command: `make docker-build docker-push`
4. After this you should see a package in your [GitHub profile](https://github.com/d4rkfella?tab=packages). Click on it and make it public, so your cluster can download the image without a secret.
5. Update notification-controller in your environment to use the image: `ghcr.io/d4rkfella/fluxcd/notification-controller:am-bearer`
6. Configure the feature and test.

Thanks in advance!

d4rkfella:
Ok, i build the image locally and tested it by intentionally making a mistake in my kustomization resources and the event was successfully dispatched to alertmanager. 
```
{"level":"info","ts":"2025-01-26T17:40:20.874Z","logger":"event-server","msg":"dispatching event","eventInvolvedObject":{"kind":"Kustomization","namespace":"flux-system","name":"cluster-apps","uid":"a38a6e6b-ce01-4c64-9593-60e6f99e62ed","apiVersion":"kustomize.toolkit.fluxcd.io/v1","resourceVersion":"2768828"},"message":"Kustomization/vault/vault dry-run failed: failed to create typed patch object (vault/vault; kustomize.toolkit.fluxcd.io/v1, Kind=Kustomization): .spec.targetNamespacee: field not declared in schema
```

Request is authenticated to oauth2-proxy successfully and send to the upstream alertmanager service.
```
172.16.0.40 - 2b9be63a-45cf-4bf5-8b2d-dfc9ec9a4caf - fluxalerts_service@darkfellanetwork.com [2025/01/26 17:40:20] alertmanager-operated.observability.svc.cluster.local:8082 POST / "/api/v2/alerts/" HTTP/1.1 "Go-http-client/1.1" 200 0 0.001
```

d4rkfella:
Does the fuzz_test need additional changes to test with token and with an empty string like how its setup for other providers that use token based authentication ? 
```
func Fuzz_AlertManager(f *testing.F) {
	f.Add("token", "update", "", "", []byte{}, []byte("{}"))
	f.Add("", "something", "", "else", []byte{}, []byte(""))

	f.Fuzz(func(t *testing.T,
		token, commitStatus, urlSuffix, summary string, seed, response []byte) {

		ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			w.Write(response)
			io.Copy(io.Discard, r.Body)
			r.Body.Close()
		}))
		defer ts.Close()

		var cert x509.CertPool
		_ = fuzz.NewConsumer(seed).GenerateStruct(&cert)

		alertmanager, err := NewAlertmanager(fmt.Sprintf("%s/%s", ts.URL, urlSuffix), "", token, &cert)
		if err != nil {
			return
		}
```
Sorry if asking stupid questions, trying to get a grasp of things :P

matheuscscp:
Sorry to bug you with this @d4rkfella, the fuzz tests are flaky at the moment, don't worry about that.

The CI has passed :grinning: 

Now the last thing before I merge this PR, I'll need you to please squash all the commits into one, with the message:

```
Add support for Bearer Token authentication to Provider alertmanager
```

d4rkfella:
Oh man, i screwed that up

matheuscscp:
No need to panic, I'm writing specific instructions for you to recover your work.

d4rkfella:
I'm sorry for wasting more of your time.

matheuscscp:
You work is in the commit `c3a8185465e3ebb5e0f346d1a0561e786c3cdc94`.

1. Reset your branch to the the Flux upstream latest commit: `git reset --hard fa7d9f2`. This will simply reset your local `main` branch to the exact same state of the `main` branch of `github.com/fluxcd/notification-controller`. If this command fails then it's because you probably need to add the Flux upstream remote to your local clone: `git remote add upstream git@github.com:fluxcd/notification-controller.git && git fetch upstream`. Then run the reset command again.
2. Cherry-pick pick the commit with your work: `git cherry-pick c3a8185465e3ebb5e0f346d1a0561e786c3cdc94`.
3. At this point, if you run `git log` you should see your commit on top of the latest commit of Flux upstream. If that's the case, now you can force push: `git push --force`

d4rkfella:
Thanks for your patience. That should be it :)

matheuscscp (approved):
Thanks for the contribution and for the patience addressing my comments :)

matheuscscp:
This will be released in the next Flux release, 2.5: https://fluxcd.io/roadmap/#v25-q1-2025

d4rkfella:
> Thanks for the contribution and for the patience addressing my comments :)

Well thank you, basically you did most of the work by giving me pretty precise guidance  .  Thanks for being patient with a fellow newbie :)

Issue #996, opened 2024-12-07
[RFC-007] Add support for GitHub App auth to `Provider` API


stefanprodan:
We should add support in the Provider for having a `.spec.secretRef` that points to a Kubernetes Secret containing GitHub App keys. When the provider `.spec.type` is set to `github` or `github-dispatch`, the controller should switch to GitHub App auth if the referred secret contains `githubApp` prefix keys.

xref: https://github.com/fluxcd/flux2/issues/4846

PR #1058, opened 2025-02-27
[RFC-007] GitHub App authentication support for github and github-dispatch providers

dipti-pai:
[RFC-007] GitHub App authentication support for github and github-dispatch providers

- Add providerOpts in notifier to configure authentication options for various providers.
- If token/password are not set to PAT, check if github app details are configured in secret and if found; authenticate using github-app by retrieving app installation token.
- If proxy is specified in the provider spec OR in the secret, configure github app authentication to fetch the installation token over the proxy.
- Add unit tests for providers.
- Update documentation describing the usage of github app authentication with the providers.

Fixes: #996


matheuscscp (commented):
Thanks @dipti-pai! I'll wait for the changes we made recently on the lib path and token cache for another pass

internal/server/event_handlers.go
```diff
@@ -314,6 +320,30 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider apiv
 				return nil, "", fmt.Errorf("failed to read headers from secret: %w", err)
 			}
 		}
+
+		// If token/password are not present in secret, check if provider authentication can be used.
+		if token == "" && password == "" {
+			switch provider.Spec.Type {
+			case apiv1beta3.GitHubDispatchProvider, apiv1beta3.GitHubProvider:
```
matheuscscp, line 327: I think this code does not follow our current structure. Right now we have generic inputs that are extracted from the secret and object fields, and that are passed into the `notifier.Factory`. I think the right way to introduce this feature would be passing the `secret.Data` map down to the factory constructor and let these two providers select this new variable in their respective constructors and parse this secret data there. It seems that this logic here is too specific to the providers and should not be placed in the event server handler.

This would also eliminate the new struct `ProviderOptions` which I think is also not in line with the current organization of the code.

This code parsing the secret and extracting the GitHub App keys from it can be placed in a single function inside a file called `github_helpers.go` or something similar in the `internal/notifier` package and used by both providers in their constructors to extract the GitHub App options.

dipti-pai: Thanks. My intention was to keep all secret parsing together, but agree with the perspective that the extracted options are provider specific and can be moved there.

internal/notifier/github.go
```diff
@@ -40,17 +37,24 @@ type GitHub struct {
...
-		return nil, err
+		if len(githubOpts) == 0 {
+			return nil, errors.New("github token or github app details must be specified")
+		}
+
+		token, err = getGitHubAppToken(githubOpts)
+		if err != nil {
+			return nil, err
+		}
 	}
```
matheuscscp, line 55: It seems to me that this entire `if len(token) == 0 {` block is exactly the same on both providers, so it looks to me that this entire block can be abstracted into a single function merging `getGitHubAppOptions()`, `getGitHubAppToken()` and everything in the middle. Let's call it `getGitHubToken(token, providerName, providerNamespace, proxyURL, secretData, tokenCache)`.

Even better, I can see even more common parts. To me it looks like the constructors of these two providers can almost entirely be merged into a single function `newGitHubClient(...)`. Then in the `github` provider we check if the `commitStatus` is empty right before creating the client.

matheuscscp (commented):
One more pass :hugs:

internal/notifier/github_helpers.go
```diff
@@ -0,0 +1,156 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return &GitHubDispatch{
+		Owner:  repoInfo[0],
+		Repo:   repoInfo[1],
+		Client: client,
+	}, nil
+}
```
matheuscscp, line 156: These two functions can be inlined in the two constructors since they are specific to each provider.

internal/notifier/github_helpers.go
```diff
@@ -0,0 +1,138 @@
...
+func getRepoInfoAndGithubClient(addr string, token string, certPool *x509.CertPool, proxyURL string, providerName string, providerNamespace string, secretData map[string][]byte, tokenCache *pkgcache.TokenCache) (repoInfo, error) {
+	var repoDetails repoInfo
+	if len(token) == 0 {
+		githubOpts, err := getGitHubAppOptions(providerName, providerNamespace, proxyURL, secretData, tokenCache)
+		if err != nil {
+			return repoDetails, err
+		}
+
+		if len(githubOpts) == 0 {
+			return repoDetails, errors.New("github token or github app details must be specified")
```
matheuscscp, line 84:
```suggestion
			return nil, errors.New("github token or github app details must be specified")
```

internal/notifier/github_helpers.go
```diff
@@ -0,0 +1,138 @@
...
+		}
+
+		client, err := authgithub.New(githubOpts...)
+		if err != nil {
+			return repoDetails, err
```
matheuscscp, line 89:
```suggestion
			return nil, err
```

internal/notifier/github_helpers.go
```diff
@@ -0,0 +1,138 @@
...
+	}
+
+	host, id, err := parseGitAddress(addr)
+	if err != nil {
+		return repoDetails, err
+	}
+
+	comp := strings.Split(id, "/")
+	if len(comp) != 2 {
+		return repoDetails, fmt.Errorf("invalid repository id %q", id)
```
matheuscscp, line 106:
```suggestion
		return nil, fmt.Errorf("invalid repository id %q", id)
```

internal/notifier/github_helpers.go
```diff
@@ -0,0 +1,138 @@
...
+		client, err = gogithub.NewClient(tc).WithEnterpriseURLs(host, host)
+		if err != nil {
+			return repoDetails, fmt.Errorf("could not create enterprise GitHub client: %v", err)
+		}
+	}
+
+	repoDetails.owner = comp[0]
+	repoDetails.repo = comp[1]
+	repoDetails.client = client
+	return repoDetails, nil
```
matheuscscp, line 137:
```suggestion
	return &repoInfo{comp[0], comp[1], client}, nil
```

matheuscscp (approved):
Thanks @dipti-pai, amazing work! Feel free to merge :hugs: 

I tested both providers in a private GitHub repo and both work:

`github` (with custom commit status)

Screenshot from 2025-03-21 12-17-16

`githubdispatch`

Screenshot from 2025-03-21 13-00-30

stefanprodan (approved):
LGTM

Thanks @dipti-pai for the great work and @matheuscscp for all the tests 🏅 🎖️

PR #1068, opened 2025-03-12
feat: support CEL string expressions for custom commit statuses in v1beta3 provider type

kathleenfrench:
Closes https://github.com/fluxcd/notification-controller/issues/589

this change was enabled by recent work to enable [RFC-0008](https://github.com/fluxcd/flux2/tree/main/rfcs/0008-custom-event-metadata-from-annotations) which allows for custom metadata to be injected in flux events.

we can leverage the flux runtime pkg `cel` library's support for compiling and evaluating strings (added in https://github.com/fluxcd/pkg/pull/882) to compute a commit status for git provider types (`github`, `gitlab`, `gitlea`, `bitbucketserver`, `bitbucket`, and `azuredevops`) that can make use of end-user provided event metadata along with standard values accessible to `event`, `provider`, and `alert`.

> CEL expresions with this feature can _only_ reference: `event`, `provider`, and `alert`.

to accomplish this, a new optional field is added to the `v1beta3` `provider` spec called `commitStatusExpr`. 

following some additional [discussions](https://github.com/fluxcd/notification-controller/issues/589#issuecomment-2717846934), it was elected to not re-add the `status` subresource to the `provider` to make way for these changes, but instead evaluate the CEL expression from the event server and forward the commit status to the relevant factory, if it exists/is applicable.

as part of this effort a few things to note:

- the notifier factory constructor has been refactored
- the `ProviderUID` notifier option has been removed from git providers in favor of `CommitStatus`, which is computed prior to creation of the notifier either via the custom expression or using the (_status quo_) default commit status generator function which combines the involved object kind, name, and provider UID.
- all git factory function signatures have been altered to sub `commitStatus` for `providerUID` -- error handling + tests have been added to ensure the status cannot be empty on creation of the factory, though this is just a failsafe measure and should not happen given if a custom expression is used and fails we drop the alert and if no custom expression is provided we use the pre-existing logic, just evaluated sooner rather than in `Post`.
- we run `combineEventMetadata()` prior to creating the notifier now 

so to recap: if the commit status expression fails to compile or evaluate we:
- ~log the error~ _removed based on pr feedback_
- ~emit a warning event~ _removed based on pr feedback_
- return an error
- drop the alert

if no commit status expression is provided and the provider is a supported git type, we generate a default commit status using event metadata and the provider UID with the same, existing logic just sooner in our handling of the event/alert. if the provider is not a git type, the commit status forwarded to `createNotifier` is an empty string.


internal/server/provider_commit_status_test.go
```diff
@@ -0,0 +1,209 @@
...
+		},
+	}
+
+	for _, tt := range validationTests {
+		t.Run(tt.expression, func(t *testing.T) {
+			g := NewWithT(t)
+			g.Expect(validateCommitStatusExpr(tt.expression)).To(MatchError(ContainSubstring(tt.wantError)))
+		})
+	}
+}
```
matheuscscp, line 69: We usually test success and error cases in a single top-level Test func branching if the error is expected or not

internal/server/event_handlers.go
```diff
@@ -251,19 +253,55 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
...
+// Otherwise, a default commit status will be generated using the Provider UID and event metadata.
+// If the provider is not a git provider, the commit status will be an empty string.
+// If the commitStatusExpr fails to compile or is invalid, an error will be returned.
+func createCommitStatus(ctx context.Context, provider apiv1beta3.Provider, event *eventv1.Event, alert *apiv1beta3.Alert) (commitStatus string, err error) {
+	if !isGitProvider(provider.Spec.Type) {
+		return "", nil
+	}
+
+	if provider.Spec.CommitStatusExpr != "" {
+		err = validateCommitStatusExpr(provider.Spec.CommitStatusExpr)
```
matheuscscp, line 287: It seems like the existence of this function doesn't add much value to the code to me, we could just call `newCommitStatusExpression()` here and pass the built `*cel.Expression` to `newCommitStatus()`, which avoids building the expression twice. Then the tests currently written for `validateCommitStatusExpr()` would become tests for `newCommitStatusExpression()`

internal/server/event_handlers.go
```diff
@@ -251,19 +253,55 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
...
+		if err != nil {
+			return "", fmt.Errorf("failed to evaluate commitStatusExpr: %w", err)
```
matheuscscp, line 289:
```suggestion
			return "", fmt.Errorf("failed to parse the spec.commitStatusExpr CEL expression: %w", err)
```

internal/server/event_handlers.go
```diff
@@ -251,19 +253,55 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
...
+		}
+
+		commitStatus, err = newCommitStatus(ctx, provider.Spec.CommitStatusExpr, event, alert, &provider)
+		if err != nil {
+			return "", fmt.Errorf("failed to evaluate commitStatusExpr: %w", err)
```
matheuscscp, line 294:
```suggestion
			return "", fmt.Errorf("failed to evaluate the spec.commitStatusExpr CEL expression for the event: %w", err)
```

internal/server/event_handlers.go
```diff
@@ -251,19 +253,55 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
...
+	// Create a copy of the event and combine event metadata
+	notification := *event.DeepCopy()
+	s.combineEventMetadata(ctx, &notification, alert)
+
+	// Create a commit status for the given provider and event, if applicable.
+	commitStatus, err := createCommitStatus(ctx, provider, &notification, alert)
 	if err != nil {
-		return nil, nil, "", 0, fmt.Errorf("failed to initialize notifier for provider '%s': %w", provider.Name, err)
+		logger.Error(err, "failed to evaluate commitStatusExpr", "expression", provider.Spec.CommitStatusExpr, "provider", provider.Name)
+		s.Eventf(&provider, corev1.EventTypeWarning, "CommitStatusExprEvalFailed", "failed to evaluate commitStatusExpr %q for provider %s: %s", provider.Spec.CommitStatusExpr, provider.Name, err)
```
matheuscscp, line 264: We don't need to log/send a k8s event here, just bubbling up the error up to the handler for loop for each Alert matching the event is enough, we log errors and send k8s events there

internal/notifier/factory.go
```diff
@@ -64,42 +64,92 @@ type notifierMap map[string]factoryFunc
...
-	proxy string,
-	username string,
-	channel string,
-	token string,
-	headers map[string]string,
-	certPool *x509.CertPool,
-	password string,
-	providerUID string) *Factory {
+// NotifierOption represents a functional option for configuring a notifier.
+type NotifierOption func(*notifierOptions)
```
matheuscscp, line 83: Thank you so much for this refactor :heart:

internal/notifier/azure_devops.go
```diff
@@ -40,14 +40,14 @@ type azureDevOpsClient interface {
...
-	Client      azureDevOpsClient
+	Project      string
+	Repo         string
+	CommitStatus string
+	Client       azureDevOpsClient
 }
 
 // NewAzureDevOps creates and returns a new AzureDevOps notifier.
-func NewAzureDevOps(providerUID string, addr string, token string, certPool *x509.CertPool) (*AzureDevOps, error) {
+func NewAzureDevOps(commitStatus string, addr string, token string, certPool *x509.CertPool) (*AzureDevOps, error) {
```
matheuscscp, line 50: Nice :+1:

matheuscscp (commented):
Thanks very much for this contribution @kathleenfrench!

First pass :grin: Will do another one tomorrow

internal/server/provider_commit_status.go
```diff
@@ -0,0 +1,145 @@
...
+	uidParts := strings.Split(providerUID, "-")
+	id := fmt.Sprintf("%s/%s/%s", event.InvolvedObject.Kind, event.InvolvedObject.Name, uidParts[0])
+	return strings.ToLower(id)
+}
+
+// newCommitStatus evaluates the commit status expression.
+func newCommitStatus(ctx context.Context, expr string, notification *eventv1.Event, alert *apiv1beta3.Alert, provider *apiv1beta3.Provider) (string, error) {
+	celExpr, err := newCommitStatusExpression(expr)
+	if err != nil {
+		return "", err
```
matheuscscp, line 53:
```suggestion
		return "", fmt.Errorf("failed to compile expression: %w", err)
```

internal/server/provider_commit_status.go
```diff
@@ -0,0 +1,145 @@
...
+
+	var (
+		eventMap    map[string]any
+		providerMap map[string]any
+		alertMap    map[string]any
+	)
+
+	eventMap, err = eventToMap(notification)
+	if err != nil {
+		return "", err
```
matheuscscp, line 64:
```suggestion
		return "", fmt.Errorf("failed to convert event to map: %w", err)
```

internal/server/provider_commit_status.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	providerMap, err = providerToMap(provider)
+	if err != nil {
+		return "", err
```
matheuscscp, line 69:
```suggestion
		return "", fmt.Errorf("failed to convert provider object to map: %w", err)
```

internal/server/provider_commit_status.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	alertMap, err = alertToMap(alert)
+	if err != nil {
+		return "", err
```
matheuscscp, line 74:
```suggestion
		return "", fmt.Errorf("failed to convert alert object to map: %w", err)
```

internal/notifier/factory.go
matheuscscp, line 83: I think I'd call this type just `Option` since this is the most important option type of the package and `notifier.NotifierOption` "stutters" in Go idiom

matheuscscp (commented):
Amazing work here @kathleenfrench!

matheuscscp (approved):
LGTM!

Thanks very much for this contribution @kathleenfrench! :rocket:

matheuscscp:
Btw I tested this feature in a real cluster and it works:

Screenshot from 2025-03-18 19-06-11

stefanprodan (approved):
LGTM

Thanks @kathleenfrench 🥇

Issue #1083, opened 2025-03-23
Failed creating Gitea client with custom CA: x509 certificate signed by unknown authority


sdreger:
Environment Information:
OS: 'Linux 5.15.0-134-generic 145-Ubuntu SMP Wed Feb 12 20:08:39 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux'
kubernetes cluster: 'k3s version v1.32.2+k3s1 (381620ef); go version go1.23.6'
fluxCD: 'flux version 2.5.1'
Gitea: 'Version: 1.23.5'

I'm trying to use Gitea provider with self-hosted Gitea instance. My reverse proxy serves a self-signed certificate, so I should provide a custom CA (certificate authority) to the Provider manifest.
```yaml
apiVersion: notification.toolkit.fluxcd.io/v1beta3
kind: Provider
metadata:
  name: lib-manager-go
  namespace: flux-system
spec:
  type: gitea
  address: https://gitea.private.lan/user/flux-cd-k3s
  secretRef:
    name: gitea-token
  certSecretRef:
    name: flux-system
---
apiVersion: notification.toolkit.fluxcd.io/v1beta3
kind: Alert
metadata:
  name: lib-manager-go
  namespace: flux-system
spec:
  providerRef:
    name: lib-manager-go
  eventSeverity: info
  eventSources:
    - kind: Kustomization
      name: lib-manager-go
      namespace: flux-system
```

The custom CA is in the `flux-system` secret under the `ca.crt` key:
```sh
kubectl get secret -n flux-system flux-system -o jsonpath='{ .data.ca\.crt }' | base64 -d | head -n 2
-----BEGIN CERTIFICATE-----
MIIFmTCCA4GgAwIBAgIUUZb+6uFY041w6clv+DdsuOxQtm8wDQYJKoZIhvcNAQEL
```

The Gitea secret is in the `gitea-token`:
```sh
kubectl get secret -n flux-system gitea-token -o jsonpath='{ .data.token }' | base64 -d
ea<REDACTED>ae
```

When an event is produced by the `lib-manager-go` Kustomization, I see the following error in the logs:
```json
{
    "level": "error",
    "ts": "2025-03-23T12:57:54.098Z",
    "logger": "event-server",
    "msg": "failed to dispatch notification",
    "eventInvolvedObject": {
      "kind": "Kustomization",
      "namespace": "flux-system",
      "name": "lib-manager-go",
[...]

PR #1084, opened 2025-03-23
fix: pass 'certPool' to Gitea client on creation

sdreger:
It is required when a custom CA is passed, otherwise the gitea.NewClient() call will fail with the 'tls: failed to verify certificate: x509: certificate signed by unknown authority' error. Because the current version of Gitea SDK performs a call to the '/api/v1/version' endpoint during a new client creation, so the 'certPool' must be passed when creating the client.

Resolves: #1083

### DIFF
--- a/internal/notifier/gitea.go
+++ b/internal/notifier/gitea.go
@@ -69,18 +69,16 @@ func NewGitea(commitStatus string, addr string, token string, certPool *x509.Cer
 		return nil, fmt.Errorf("invalid repository id %q", id)
 	}
 
-	client, err := gitea.NewClient(host, gitea.SetToken(token))
-	if err != nil {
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @sdreger 🏅 

PS. I realised that we're missing the proxy setting for Gitea. The `NewGitea` function should have a `proxy string` arg and when we create the transport we should be setting the proxy if one is specified like we do for all the providers, for example [here](https://github.com/fluxcd/notification-controller/blob/d62be26a3f427e51ed70b2714f8c480b2411f049/internal/notifier/factory.go#L212). @sdreger if you have time please contribute add the proxy config in a PR. Thanks again!

PR #1086, opened 2025-03-25
Fix Slack `chat.postMessage` error handling

### LINKED ISSUE #1048: Slack `chat.postMessage` errors are not caught
`postMessage` function assumes that HTTP responses have a meaningful status code (2xx for success, 4xx or 5xx for error).
https://github.com/fluxcd/notification-controller/blob/2763e548bd16298f13739108f7635a6438731b96/internal/notifier/client.go#L103

Slack `chat.postMessage` API does not satisfy this assumption.
https://api.slack.com/methods/chat.postMessage
Seems that it always responds with `200 OK`.
`ok` and `error` fields in response body JSON indicate an error, if any.

Thus, Slack errors are not caught by `postMessage` function, and will not appear in notification-controller logs nor Alert resource Events.

ordovicia:
Fixes #1048 

This pull request fixes the error handling of Slack `chat.postMessage` API.
The API always returns `200 OK`, so we need to inspect the response body indicating whether the request succeeded or not.

This pull request is still in draft -- I am going to add a test for this change.
But before that, I would like some comments on whether this implementation direction is good.

### DIFF
--- a/internal/notifier/alertmanager.go
+++ b/internal/notifier/alertmanager.go
@@ -137,13 +137,17 @@ func (s *Alertmanager) Post(ctx context.Context, event eventv1.Event) error {
 		},
 	}
 
-	var opts []requestOptFunc
-	if s.Token != "" {
-		opts = append(opts, func(request *retryablehttp.Request) {
-			request.Header.Add("Authorization", "Bearer "+s.Token)
-		})
+	opts := []messageOption{
+		withProxy(s.ProxyURL),
+		withCertPool(s.CertPool),
 	}
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, opts...)
+	if s.Token != "" {
+		opts = append(opts,
+			withRequestOption(func(request *retryablehttp.Request) {
+				request.Header.Add("Authorization", "Bearer "+s.Token)
+			}))
+	}
+	err := postMessage(ctx, s.URL, payload, opts...)
 	if err != nil {
 		return fmt.Errorf("postMessage failed: %w", err)
 	}
--- a/internal/notifier/client.go
+++ b/internal/notifier/client.go
@@ -33,16 +33,25 @@ import (
 )
 
 type messageOptions struct {
-	proxy    string
-	certPool *x509.CertPool
-	reqOpts  []requestOptFunc
+	proxy            string
+	certPool         *x509.CertPool
+	reqOpts          []requestOptFunc
+	validateResponse func(*http.Response) bool
 }
 type requestOptFunc func(*retryablehttp.Request)
 
 type messageOption func(*messageOptions)
 
 func postMessage(ctx context.Context, address string, payload interface{}, options ...messageOption) error {
-	opts := &messageOptions{}
+	opts := &messageOptions{
+		// Default validateResponse function varifies that the response status code is 200, 202 or 201.
+		validateResponse: func(resp *http.Response) bool {
+			return resp.StatusCode == http.StatusOK ||
+				resp.StatusCode == http.StatusAccepted ||
+				resp.StatusCode == http.StatusCreated
+		},
+	}
+
 	for _, o := range options {
 		o(opts)
 	}
@@ -61,6 +70,7 @@ func postMessage(ctx context.Context, address string, payload interface{}, optio
 	if err != nil {
 		return fmt.Errorf("failed to create a new request: %w", err)
 	}
+
 	if ctx != nil {
 		req = req.WithContext(ctx)
 	}
@@ -68,12 +78,14 @@ func postMessage(ctx context.Context, address string, payload interface{}, optio
 	for _, o := range opts.reqOpts {
 		o(req)
 	}
+
 	resp, err := httpClient.Do(req)
 	if err != nil {
 		return fmt.Errorf("failed to execute request: %w", err)
 	}
+	defer resp.Body.Close()
 
-	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusCreated {
+	if !opts.validateResponse(resp) {
 		b, err := io.ReadAll(resp.Body)
 		if err != nil {
 			return fmt.Errorf("unable to read response body, %s", err)
@@ -99,6 +111,12 @@ func withCertPool(certPool *x509.CertPool) messageOption {
 func withRequestOption(reqOpt requestOptFunc) messageOption {
 	return func(opts *messageOptions) {
 		opts.reqOpts = append(opts.reqOpts, reqOpt)
+	}
+}
+
+func withValidateResponse(validateResponse func(*http.Response) bool) messageOption {
+	return func(opts *messageOptions) {
+		opts.validateResponse = validateResponse
 	}
 }
 
--- a/internal/notifier/client.go
+++ b/internal/notifier/client.go
@@ -32,29 +32,99 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
+type messageOptions struct {
+	proxy    string
+	certPool *x509.CertPool
+	reqOpts  []requestOptFunc
+}
 type requestOptFunc func(*retryablehttp.Request)
 
-func postMessage(ctx context.Context, address, proxy string, certPool *x509.CertPool, payload interface{}, reqOpts ...requestOptFunc) error {
+type messageOption func(*messageOptions)
+
+func postMessage(ctx context.Context, address string, payload interface{}, options ...messageOption) error {
+	opts := &messageOptions{}
+	for _, o := range options {
+		o(opts)
+	}
+
+	httpClient, err := newHTTPClient(opts)
+	if err != nil {
+		return err
+	}
+
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return fmt.Errorf("marshalling notification payload failed: %w", err)
+	}
+
+	req, err := retryablehttp.NewRequest(http.MethodPost, address, data)
+	if err != nil {
+		return fmt.Errorf("failed to create a new request: %w", err)
+	}
+	if ctx != nil {
+		req = req.WithContext(ctx)
+	}
+	req.Header.Set("Content-Type", "application/json")
+	for _, o := range opts.reqOpts {
+		o(req)
+	}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to execute request: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK && resp.StatusCode != http.StatusAccepted && resp.StatusCode != http.StatusCreated {
+		b, err := io.ReadAll(resp.Body)
+		if err != nil {
+			return fmt.Errorf("unable to read response body, %s", err)
+		}
+		return fmt.Errorf("request failed with status code %d, %s", resp.StatusCode, string(b))
+	}
+
+	return nil
+}
+
+func withProxy(proxy string) messageOption {
+	return func(opts *messageOptions) {
+		opts.proxy = proxy
+	}
+}
+
+func withCertPool(certPool *x509.CertPool) messageOption {
+	return func(opts *messageOptions) {
+		opts.certPool = certPool
+	}
+}
+
+func withRequestOption(reqOpt requestOptFunc) messageOption {
+	return func(opts *messageOptions) {
+		opts.reqOpts = append(opts.reqOpts, reqOpt)
+	}
+}
+
+func newHTTPClient(opts *messageOptions) (*retryablehttp.Client, error) {
 	httpClient := retryablehttp.NewClient()
-	if certPool != nil {
+	if opts.certPool != nil {
 		httpClient.HTTPClient.Transport = &http.Transport{
 			TLSClientConfig: &tls.Config{
-				RootCAs: certPool,
+				RootCAs: opts.certPool,
 			},
 		}
 	}
 
-	if proxy != "" {
-		proxyURL, err := url.Parse(proxy)
+	if opts.proxy != "" {
+		proxyURL, err := url.Parse(opts.proxy)
 		if err != nil {
-			return fmt.Errorf("unable to parse proxy URL '%s', error: %w", proxy, err)
+			return nil, fmt.Errorf("unable to parse proxy URL '%s', error: %w", opts.proxy, err)
 		}
+
 		var tlsConfig *tls.Config
-		if certPool != nil {
+		if opts.certPool != nil {
 			tlsConfig = &tls.Config{
-				RootCAs: certPool,
+				RootCAs: opts.certPool,
 			}
 		}
+
 		httpClient.HTTPClient.Transport = &http.Transport{
 			Proxy:           http.ProxyURL(proxyURL),
 			TLSClientConfig: tlsConfig,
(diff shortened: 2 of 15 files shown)


matheuscscp:
Hi @ordovicia, is there an alternative where we can make less changes, and preferably touch only the slack provider?

ordovicia:
Thank you for the comment!
I come up with two alternative:

### 1. Add specialized `postMessage` for Slack provider

We could copy the existing `postMessage` and add specialized error handling for Slack provider.
Other providers are left untouched and continue to use existing `postMessage`.

However, this approach would likely be less maintainable since similar code would exist in two places.

### 2. Add an argument to `postMessage`

Instead of fully transforming `postMessage` into the functional option pattern, we could add an argument to `postMessage` like this:

```go
func postMessage(
  ctx context.Context, address, proxy string, certPool *x509.CertPool, payload interface{},
  validateResponse func(*http.Response) bool, // Added
  reqOpts ...requestOptFunc,
) error {
```

Slack provider will set `validateResponse` to customize the validation of HTTP responses.
Other providers will just pass `nil`, indicating that no customized validation is required.

But, in general, functions should not have too many arguments.
I took the functional option pattern to limit the number of arguments in the function signature and thus keep the readability.

matheuscscp:
Do you see any issues with the approach I described here?

https://github.com/fluxcd/notification-controller/issues/1048#issuecomment-2667921448

ordovicia:
`reqOpts` is for modifying requests, so we need another option to customize response validation for this fix.
Straightforward approach is adding an argument to `postMessage` (approach 2 in https://github.com/fluxcd/notification-controller/pull/1086#issuecomment-2792676180), but this could harm the maintainability.
On the other hand, aggregating options into a struct seems good.
So I updated this PR so that `postMessage` accepts an option struct directly, instead of introducing the functional option pattern.
What do you think?

internal/notifier/client.go
```diff
@@ -32,29 +32,90 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
-type requestOptFunc func(*retryablehttp.Request)
+type postOption struct {
+	proxy             string
+	certPool          *x509.CertPool
+	requestModifiers  []requestModifier
```
matheuscscp, line 38: I don't think this option needs to be a slice, all the callers are passing only a single function, and that really seems to be enough, I don't see why we would pass more than one function for this

internal/notifier/client.go
```diff
@@ -32,29 +32,90 @@ import (
...
+	responseValidator func(*http.Response) bool
```
matheuscscp, line 39: Validators should return an `error`, not a boolean. Maybe the response body has information about the error and we need to bubble it up so the error gets logged

internal/notifier/client.go
```diff
@@ -32,29 +32,90 @@ import (
...
+}
+type requestModifier func(*retryablehttp.Request)
```
matheuscscp, line 41: Either define types for both the request modifier and the response validator, or don't define those types. I'm fine with not defining either

internal/notifier/client.go
```diff
@@ -32,29 +32,90 @@ import (
 	"github.com/hashicorp/go-retryablehttp"
 )
 
-type requestOptFunc func(*retryablehttp.Request)
+type postOption struct {
```
matheuscscp, line 35: Let's rename this to `postOptions`

internal/notifier/client.go
```diff
@@ -32,29 +32,90 @@ import (
...
+	proxy             string
+	certPool          *x509.CertPool
+	requestModifiers  []requestModifier
+	responseValidator func(*http.Response) bool
+}
+type requestModifier func(*retryablehttp.Request)
+
+func postMessage(ctx context.Context, address string, payload interface{}, opt *postOption) error {
```
matheuscscp, line 43: I still prefer receiving the options as a variadic slice of functional options:

```suggestion
func postMessage(ctx context.Context, address string, payload interface{}, opts ...postOptions) error {
```

Where:

```go
type postOption func(*postOptions)
```

Let's define also `with<Something>()` helper functions for passing the options inside the providers. We recently refactor the factory constructor to adhere to this style, let's follow it everywhere :pray:

matheuscscp (commented):
Hi @ordovicia, sorry for the slowness, still catching up to many things after KubeCon EU last week.

I remember now what we're doing here, sorry for this comment:

> and preferably touch only the slack provider?

To fix this properly we're indeed gonna need to touch several providers, but it's for the best.

ordovicia:
Thank you for the through review!
I updated the pull request according to your comments

internal/notifier/client.go
```diff
@@ -32,29 +32,117 @@ import (
...
+	options := &postOptions{
+		// Default validateResponse function verifies that the response status code is 200, 202 or 201.
+		responseValidator: func(resp *http.Response) error {
+			if resp.StatusCode == http.StatusOK ||
+				resp.StatusCode == http.StatusAccepted ||
+				resp.StatusCode == http.StatusCreated {
+				return nil
+			}
+
+			return fmt.Errorf("request failed with status %s", resp.Status)
```
matheuscscp, line 54: In the previous version of the default validation we were also reading the body and returning it in the error.

internal/notifier/alertmanager.go
```diff
@@ -137,15 +137,19 @@ func (s *Alertmanager) Post(ctx context.Context, event eventv1.Event) error {
 		},
 	}
 
-	var opts []requestOptFunc
+	opts := []postOption{
+		withProxy(s.ProxyURL),
```
matheuscscp, line 141: There's no point in making options optional if we always pass them. We need to check if the option is set and only pass it if it is.

internal/notifier/alertmanager.go
```diff
@@ -137,15 +137,19 @@ func (s *Alertmanager) Post(ctx context.Context, event eventv1.Event) error {
...
+		withCertPool(s.CertPool),
```
matheuscscp, line 142: Same here.

internal/notifier/discord.go
```diff
@@ -90,8 +90,7 @@ func (s *Discord) Post(ctx context.Context, event eventv1.Event) error {
 
 	payload.Attachments = []SlackAttachment{a}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, nil, payload)
-	if err != nil {
+	if err := postMessage(ctx, s.URL, payload, withProxy(s.ProxyURL)); err != nil {
```
matheuscscp, line 93: And here.

internal/notifier/forwarder.go
```diff
@@ -77,18 +77,25 @@ func (f *Forwarder) Post(ctx context.Context, event eventv1.Event) error {
...
-		}
-	})
 
-	if err != nil {
+	if err := postMessage(
+		ctx,
+		f.URL,
+		event,
+		withProxy(f.ProxyURL),
+		withCertPool(f.CertPool),
```
matheuscscp, line 86: And here.

internal/notifier/google_chat.go
```diff
@@ -142,8 +142,7 @@ func (s *GoogleChat) Post(ctx context.Context, event eventv1.Event) error {
 		Cards: []GoogleChatCard{card},
 	}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, nil, payload)
-	if err != nil {
+	if err := postMessage(ctx, s.URL, payload, withProxy(s.ProxyURL)); err != nil {
```
matheuscscp, line 145: And here.

internal/notifier/grafana.go
```diff
@@ -84,15 +84,21 @@ func (g *Grafana) Post(ctx context.Context, event eventv1.Event) error {
...
-			request.Header.Add("Authorization", "Bearer "+g.Token)
-		}
-	})
-	if err != nil {
+	if err := postMessage(
+		ctx,
+		g.URL,
+		payload,
+		withProxy(g.ProxyURL),
+		withCertPool(g.CertPool),
```
matheuscscp, line 92: And here.

internal/notifier/matrix.go
```diff
@@ -65,11 +65,16 @@ func (m *Matrix) Post(ctx context.Context, event eventv1.Event) error {
...
-	err = postMessage(ctx, fullURL, "", m.CertPool, payload, func(request *retryablehttp.Request) {
-		request.Method = http.MethodPut
-		request.Header.Add("Authorization", "Bearer "+m.Token)
-	})
-	if err != nil {
+	if err := postMessage(
+		ctx,
+		fullURL,
+		payload,
+		withCertPool(m.CertPool),
```
matheuscscp, line 72: And here.

internal/notifier/opsgenie.go
```diff
@@ -78,12 +78,18 @@ func (s *Opsgenie) Post(ctx context.Context, event eventv1.Event) error {
...
-		req.Header.Set("Authorization", "GenieKey "+s.ApiKey)
-	})
-
-	if err != nil {
+	if err := postMessage(
+		ctx,
+		s.URL,
+		payload,
+		withProxy(s.ProxyURL),
+		withCertPool(s.CertPool),
```
matheuscscp, line 86: And here.

internal/notifier/pagerduty.go
```diff
@@ -54,19 +54,30 @@ func (p *PagerDuty) Post(ctx context.Context, event eventv1.Event) error {
...
-	e := toPagerDutyV2Event(event, p.RoutingKey)
-	err := postMessage(ctx, p.Endpoint+"/v2/enqueue", p.ProxyURL, p.CertPool, e)
-	if err != nil {
+
+	if err := postMessage(
+		ctx,
+		p.Endpoint+"/v2/enqueue",
+		toPagerDutyV2Event(event, p.RoutingKey),
+		withProxy(p.ProxyURL),
+		withCertPool(p.CertPool),
```
matheuscscp, line 63: And here.

internal/notifier/pagerduty.go
```diff
@@ -54,19 +54,30 @@ func (p *PagerDuty) Post(ctx context.Context, event eventv1.Event) error {
...
 	if event.Severity == eventv1.EventSeverityInfo {
-		ce := toPagerDutyChangeEvent(event, p.RoutingKey)
-		err = postMessage(ctx, p.Endpoint+"/v2/change/enqueue", p.ProxyURL, p.CertPool, ce)
-		if err != nil {
+		if err := postMessage(
+			ctx,
+			p.Endpoint+"/v2/change/enqueue",
+			toPagerDutyChangeEvent(event, p.RoutingKey),
+			withProxy(p.ProxyURL),
+			withCertPool(p.CertPool),
```
matheuscscp, line 75: And here.

internal/notifier/rocket.go
```diff
@@ -83,8 +83,7 @@ func (s *Rocket) Post(ctx context.Context, event eventv1.Event) error {
 
 	payload.Attachments = []SlackAttachment{a}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload)
-	if err != nil {
+	if err := postMessage(ctx, s.URL, payload, withProxy(s.ProxyURL), withCertPool(s.CertPool)); err != nil {
```
matheuscscp, line 86: And here.

internal/notifier/slack.go
```diff
@@ -118,13 +122,52 @@ func (s *Slack) Post(ctx context.Context, event eventv1.Event) error {
...
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, func(request *retryablehttp.Request) {
-		if s.Token != "" {
-			request.Header.Add("Authorization", "Bearer "+s.Token)
-		}
-	})
-	if err != nil {
+	opts := []postOption{
+		withProxy(s.ProxyURL),
+		withCertPool(s.CertPool),
```
matheuscscp, line 127: And here.

internal/notifier/teams.go
```diff
@@ -161,8 +161,7 @@ func (s *MSTeams) Post(ctx context.Context, event eventv1.Event) error {
 		payload = buildMSTeamsAdaptiveCardPayload(&event, objName)
 	}
 
-	err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload)
-	if err != nil {
+	if err := postMessage(ctx, s.URL, payload, withProxy(s.ProxyURL), withCertPool(s.CertPool)); err != nil {
```
matheuscscp, line 164: And here.

internal/notifier/webex.go
```diff
@@ -108,10 +108,18 @@ func (s *Webex) Post(ctx context.Context, event eventv1.Event) error {
...
 
-	if err := postMessage(ctx, s.URL, s.ProxyURL, s.CertPool, payload, func(request *retryablehttp.Request) {
-		request.Header.Add("Authorization", "Bearer "+s.Token)
-	}); err != nil {
+	if err := postMessage(
+		ctx,
+		s.URL,
+		payload,
+		withProxy(s.ProxyURL),
+		withCertPool(s.CertPool),
```
matheuscscp, line 116: And here.

matheuscscp (commented):
Another pass, this one is mostly about how we are always passing options, when we should be doing it conditionally like this:

https://github.com/fluxcd/notification-controller/blob/e970526b1e569dedcdb356b4818162a5239a7af0/internal/server/event_handlers.go#L386-L434

internal/notifier/client.go
ordovicia: Actually the body was reported in the error handling around L93, but it was not clear.
I refactored error handling in 900ea8cfdfb35a1dd183da2235eb7109aaff3867

ordovicia:
Hi @matheuscscp could you take a look again? Thank you!

matheuscscp:
Thanks @ordovicia, I'll have time to review tomorrow :ok_hand:

ordovicia:
No problem! I will add test cases if the direction of implementation is ok

internal/notifier/slack.go
```diff
@@ -118,13 +122,56 @@ func (s *Slack) Post(ctx context.Context, event eventv1.Event) error {
...
+		return fmt.Errorf("unable to read response body: %w", err)
+	}
+	resp.Body = io.NopCloser(bytes.NewBuffer(body))
+
+	type slackResponse struct {
+		Ok    bool   `json:"ok"`
+		Error string `json:"error"`
+	}
+	slackResp := slackResponse{}
+	if err := json.Unmarshal(body, &resp); err != nil {
```
matheuscscp, line 169: This defer will close the new body you assigned below, and not the original body which is what you want. You don't need to defer this because there's already a defer outside in case reading errors out. You only need to close the original body reader if the ReadAll() succeeds.

However, since reading the body in the response validator is already starting to get messed up, I think a much better idea is reading the body before calling the response validator and passing just a `[]byte` to the validator instead.

internal/notifier/slack.go
```diff
@@ -118,13 +122,56 @@ func (s *Slack) Post(ctx context.Context, event eventv1.Event) error {
...
 	return nil
 }
+
+// validateResponse validates that a chat.postMessage API response is successful.
+// chat.postMessage API always returns 200 OK.
+// See https://api.slack.com/methods/chat.postMessage.
+//
+// On the other hand, incoming webhooks return more expressive HTTP status codes.
+// See https://api.slack.com/messaging/webhooks#handling_errors.
+func (s *Slack) validateResponse(resp *http.Response) error {
```
matheuscscp, line 155: Please build notification-controller and test it manually with a real Slack Workspace, both on a successful and unsuccessful response. I suppose using a wrong bearer token will do the trick for making the request unsuccessful? Please post evidence of your manual tests as a comment in the PR, e.g. screenshots, logs, etc.

matheuscscp:
Please also write unit tests for the client, no need to worry about writing tests for the particular providers.

internal/notifier/alertmanager.go
matheuscscp: Looks like you understood it, basically I was talking about the `if` statements for adding `withSomething(...)` :+1:

matheuscscp:
@ordovicia We will start releasing the controllers for Flux 2.6 on May 26 and finish the release in May 29, will you have time to finish working on this by then?

internal/notifier/slack.go
ordovicia: > I think a much better idea is reading the body before calling the response validator and passing just a `[]byte` to the validator instead.

Surely that's a better idea! I rewrote the response validator in that way.
Ideally we should read the response body only when it's necessary for e.g. error reporting, but we can do more optimization in another pull request.

ordovicia:
Sorry for my late response. I will try to write unit tests and do manual tests in a few days

internal/notifier/slack.go
```diff
@@ -118,13 +119,49 @@ func (s *Slack) Post(ctx context.Context, event eventv1.Event) error {
...
+		}),
+	}
+	if s.ProxyURL != "" {
+		opts = append(opts, withProxy(s.ProxyURL))
+	}
+	if s.CertPool != nil {
+		opts = append(opts, withCertPool(s.CertPool))
+	}
+	if s.URL == "https://slack.com/api/chat.postMessage" {
+		opts = append(opts, withResponseValidator(s.validateResponse))
```
matheuscscp, line 136:
```suggestion
		opts = append(opts, withResponseValidator(validateResponse))
```

ordovicia:
I added some unit tests. I will do a manual test for a real Slack workspace soon!

matheuscscp:
LGTM! @ordovicia After testing, if no more changes are needed (it works), please rebase and squash all commits into one. Thanks!

PR #1087, opened 2025-03-25
feat: add 'proxyURL' support to Gitea notifier

sdreger:
As was mentioned [here](https://github.com/fluxcd/notification-controller/pull/1084#pullrequestreview-2709351090) we're missing the proxy setting for Gitea. This PR introduces support for Gitea 'proxyURL'.

### DIFF
--- a/internal/notifier/gitea.go
+++ b/internal/notifier/gitea.go
@@ -45,7 +45,7 @@ type Gitea struct {
 
 var _ Interface = &Gitea{}
 
-func NewGitea(commitStatus string, addr string, token string, certPool *x509.CertPool) (*Gitea, error) {
+func NewGitea(commitStatus string, addr string, proxyURL string, token string, certPool *x509.CertPool) (*Gitea, error) {
 	if len(token) == 0 {
 		return nil, errors.New("gitea token cannot be empty")
 	}
@@ -74,6 +74,14 @@ func NewGitea(commitStatus string, addr string, token string, certPool *x509.Cer
 		tr.TLSClientConfig = &tls.Config{
 			RootCAs: certPool,
 		}
+	}
+
+	if proxyURL != "" {
+		parsedProxyURL, err := url.Parse(proxyURL)
+		if err != nil {
+			return nil, errors.New("invalid proxy URL")
+		}
+		tr.Proxy = http.ProxyURL(parsedProxyURL)
 	}
 
 	client, err := gitea.NewClient(host, gitea.SetToken(token), gitea.SetHTTPClient(&http.Client{Transport: tr}))
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/cdevents/sdk-go v0.4.1
 	github.com/chainguard-dev/git-urls v1.0.2
 	github.com/containrrr/shoutrrr v0.8.0
+	github.com/elazarl/goproxy v1.7.2
 	github.com/fluxcd/cli-utils v0.36.0-flux.12
 	github.com/fluxcd/notification-controller/api v1.5.0
 	github.com/fluxcd/pkg/apis/event v0.16.0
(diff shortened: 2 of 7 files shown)


sdreger:
I'm not sure about `testHTTPProxyServer`. Right now it's only used by `gitea_test.go`, but I decided to put it in a separate file `proxy_server_test.go`, since it can be used in other integration tests designed to test interaction with a proxy server.

stefanprodan:
For proxy testing in source-controller we used this https://github.com/fluxcd/source-controller/blob/414b7dbf9552dd35e1f4dfb4aa6862f2bc65ea08/tests/proxy/proxy.go#L23

matheuscscp:
I introduced those `tests/listener` and `tests/proxy` in SC and am wondering if we should move those to `fluxcd/pkg` now that we want to use this here in NC too :thinking: I think you can just copy the same structure here for now, we'll think about this later

sdreger:
I removed my own proxy implementation and switched to using the `github.com/elazarl/goproxy` for proxy-related tests.

I also copied the: https://github.com/fluxcd/source-controller/tree/main/tests/proxy and https://github.com/fluxcd/source-controller/tree/main/tests/listener packages into the project as suggested above.

The PR is ready for review.

internal/notifier/gitea.go
```diff
@@ -76,6 +76,14 @@ func NewGitea(commitStatus string, addr string, token string, certPool *x509.Cer
 		}
 	}
 
+	if proxyURL != "" {
+		parsedProxyURL, err := url.Parse(proxyURL)
+		if err != nil {
+			return nil, fmt.Errorf("unable to parse proxy URL '%s', error: %w", proxyURL, err)
```
stefanprodan, line 82:
```suggestion
			return nil, errors.New("invalid proxy URL")
```
We don't want to log the proxy as it can contain user/pass in the URL.

matheuscscp (approved):
Looks like you need to update a test to match the last change. Aside from that, LGTM! Please test this PR and post here some evidence of the test, e.g. logs/screenshots/something

sdreger:
Rebased the branch and squashed the commits.

---
Providing evidence of the test:
1. I've built and pushed a custom image to my personal registry: https://hub.docker.com/r/sdreger/flux-notification-controller
2. Then I've disabled `flux-system` reconciliation: `flux suspend kustomization flux-system`.
3. After that I've deployed the custom image to my cluster with: `make docker-deploy`.
4. Now `notification-controller` runs the custom image:
```sh
kubectl -n flux-system get deployment/notification-controller -o jsonpath="{.spec.template.spec.containers[0].image}"

sdreger/flux-notification-controller:latest
```

Here is my personal Gitea provider configuration:
```yaml
kind: Provider
metadata:
  name: lib-manager-go
  namespace: flux-system
spec:
  type: gitea
  address: https://gitea.*****.lan/sdreger/lib-manager-go
  secretRef:
    name: gitea-token
  certSecretRef:
    name: flux-system
  proxy: http://192.168.50.220:9999
```

Here is [proxy](https://github.com/jthomperoo/simple-proxy) running:
```sh
simple-proxy --port 9999 -v 1
I0409 16:02:09.941260 1108523 main.go:117] Starting HTTP proxy...

I0409 16:11:37.595771 1108523 proxy.go:37] Serving 'CONNECT' request from '10.42.0.100:44534' to 'gitea.*****.lan:443'
```

Logs from the `notification-controller`:
```json
{
    "level": "info",
    "ts": "2025-04-09T13:11:37.592Z",
    "logger": "event-server",
    "msg": "dispatching event",
    "eventInvolvedObject": {
      "kind": "Kustomization",
      "namespace": "flux-system",
      "name": "lib-manager-go",
      "uid": "54a287a0-68cc-4aa1-95f0-3b7bd0d68163",
      "apiVersion": "kustomize.toolkit.fluxcd.io/v1",
      "resourceVersion": "3080619"
    },
    "message": "Reconciliation finished in 160.03991ms, next run in 5m0s"
  }
```

And the final result (green checkmark in my Gitea interface):
image

PR #1112, opened 2025-05-05
Reintroduce default state machine for Provider controller

matheuscscp:
After introducing the token cache the Provider controller now has a reason to keep a finalizer at all times and use it in order to cleanup the token cache when a Provider object is deleted. So instead of being a controller that tries to remove the finalizer all the time, it's now again a controller that tries to add the finalizer all the time until deletion is observed.

### DIFF
--- a/internal/controller/provider_controller_test.go
+++ b/internal/controller/provider_controller_test.go
@@ -52,61 +52,29 @@ func TestProviderReconciler(t *testing.T) {
 	}
 	providerKey := client.ObjectKeyFromObject(provider)
 
-	// Remove finalizer at create.
-
-	provider.ObjectMeta.Finalizers = append(provider.ObjectMeta.Finalizers, "foo.bar", apiv1.NotificationFinalizer)
+	// Create without finalizer.
 	provider.Spec = apiv1beta3.ProviderSpec{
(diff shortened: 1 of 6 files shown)


internal/controller/provider_controller_test.go
```diff
@@ -52,61 +52,29 @@ func TestProviderReconciler(t *testing.T) {
...
-	g.Expect(err).ToNot(HaveOccurred())
-
-	// Suspend the provider to prevent finalizer from getting removed.
-	// Ensure only flux finalizer is set to allow the object to be garbage
-	// collected at the end.
-	// NOTE: Suspending and updating finalizers are done separately here as
-	// doing them in a single patch results in flaky test where the finalizer
-	// update doesn't gets registered with the kube-apiserver, resulting in
-	// timeout waiting for finalizer to appear on the object below.
-	provider.Spec.Suspend = true
```
matheuscscp, line 92: If in the previous version .suspend blocks removing the finalizer, in this PR it should block adding the finalizer. But we don't do this anywhere else, so I think .suspend is not needed for anything in this controller.

stefanprodan (approved):
LGTM

PR #1133, opened 2025-06-26
Add ProxySecretRef field to Provider API

### LINKED ISSUE #1131: Deprecate `spec.proxy` and introduce `spec.proxySecretRef` for the Provider API
xref: https://github.com/fluxcd/pkg/issues/949

As discussed in [this](https://cloud-native.slack.com/archives/C01A402P0R2/p1750448321600989) CNCF Slack thread, we decided to deprecate `spec.proxy` in Provider `v1beta3` and remove it in `v1`.

The replacement will be the widespread `spec.proxySecretRef`.

For Flux 2.7 we need to introduce `spec.proxySecretRef` and print an error log about the deprecation and removal in Provider v1 GA when either `spec.proxy` is used or the `proxy` key is present in the Secret from `spec.secretRef`.

cappyzawa:
Resolves #1131 by introducing `spec.proxySecretRef` to the Provider API for secure proxy configuration through dedicated Secrets.


cappyzawa:
Fixed via [40a2f0e](https://github.com/fluxcd/notification-controller/pull/1133/commits/40a2f0e824fc9dbf81cff6104fad22209325bb65) .

internal/server/event_handlers.go
```diff
@@ -336,11 +338,12 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider *api
...
 
 		if val, ok := secret.Data["proxy"]; ok {
-			proxy = strings.TrimSpace(string(val))
-			_, err := url.Parse(proxy)
+			deprecatedProxy = strings.TrimSpace(string(val))
+			_, err := url.Parse(deprecatedProxy)
 			if err != nil {
 				return nil, "", fmt.Errorf("invalid 'proxy' in secret '%s/%s'", secret.Namespace, secret.Name)
 			}
+			log.FromContext(ctx).Error(nil, "warning: specifying proxy with 'proxy' key in secret is deprecated, use ProxySecretRef with 'address' key instead. This will be removed in v1.")
```
matheuscscp, line 346: I know we are logging about `spec.proxy` in the controller, but we should probably log about it while handling events as well

matheuscscp: Agreed!

matheuscscp: Actually, I think we should even remove the log we added on the controller

cappyzawa: [4cc3181](https://github.com/fluxcd/notification-controller/pull/1133/commits/4cc31810dc9de2f911544d8d17963a567c0eb2a5)

internal/server/event_handlers.go
```diff
@@ -311,7 +312,11 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider *api
 
 	webhook := provider.Spec.Address
 	username := provider.Spec.Username
-	proxy := provider.Spec.Proxy
+	// TODO: Remove deprecated proxy handling when Provider v1 is released.
+	deprecatedProxy := provider.Spec.Proxy
+	if deprecatedProxy != "" {
+		log.FromContext(ctx).Error(nil, "spec.proxy is deprecated, please use spec.proxySecretRef instead. This field will be removed in v1.")
```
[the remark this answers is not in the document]
matheuscscp: Edit: Added `warning: ` as a prefix

matheuscscp (approved):
LGTM! Thanks very much @cappyzawa, pls squash so we can merge

Let's file a follow up issue about this PR: Right now we are parsing the proxy URL, serializing it, then parsing again in the notifiers. This should be simplified to parse only once, in the lib. The `*url.URL` should be wired down to all notifiers. This can be done when we remove the deprecated fields a few good months from now, for Flux 2.8.

cappyzawa:
@matheuscscp Got it! That’s a great refactoring idea!

PR #1139, opened 2025-07-06
Unify BasicAuth processing using pkg/runtime/secrets

cappyzawa:
## Summary

This PR refactors the createNotifier function to use `pkg/runtime/secrets.BasicAuthFromSecret` for standardized BasicAuth handling while maintaining token-first authentication precedence as part of https://github.com/fluxcd/flux2/issues/5433.

- Extract notification-controller specific Secret fields (address, proxy, token, headers) into dedicated `extractAuthFromSecret` function
- Integrate `secrets.BasicAuthFromSecret` for username/password processing  
- Maintain existing authentication priority: token → BasicAuth fallback
- Optimize Secret reading from multiple calls to single call per provider
- Preserve full backward compatibility with existing Secret formats

## Follow-up Work

- **Proxy URL parsing optimization**: Simplify proxy URL handling to parse only once in the lib and wire `*url.URL` down to all notifiers (planned for Flux v1 when deprecated fields are removed)
- **Git-based notifiers mTLS support**: Extend mTLS capabilities to GitHub, GitLab, Gitea, Bitbucket, and Azure DevOps notifiers

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -18,7 +18,6 @@ package server
 
 import (
 	"context"
-	"crypto/tls"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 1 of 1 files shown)


stefanprodan (approved):
LGTM

Thanks @cappyzawa 🏅 

PS. Please rebase

matheuscscp:
We don't cache Secrets by default so this change will have a significant impact:

https://github.com/fluxcd/notification-controller/blob/6e18c487d7c952099d9f3b04067c26f4b1ecbfce/internal/features/features.go#L35-L39

This is an example of what I described here:

https://github.com/fluxcd/flux2/issues/5433#issuecomment-3031563739

@cappyzawa Can you please file a PR to fix this? Feel free to choose between the two alternatives I described in the comment linked above

PR #1140, opened 2025-07-07
Add proxy support to Telegram notifier

### LINKED ISSUE #432: Proxy in provider spec
There is a proxy parameter in the spec. But not all types support it. For example, telegram does not support [link.](https://github.com/fluxcd/notification-controller/blob/main/internal/notifier/factory.go#L89)

I spent a lot of time to understand this. Shouldn't all types support this setting? Or at least have a mention in the documentation that they don't support proxies.

cappyzawa:
## Summary

Add proxy support to Telegram notifier by replacing shoutrrr with direct Telegram Bot API calls.

## Changes

- Replace shoutrrr.Send with postMessage function for proxy support
- Add ProxyURL field to Telegram struct
- Update v1beta3 documentation to remove proxy limitation

## Testing

Proxy functionality should work with ProxySecretRef added in #1133. Implementation follows existing patterns and should work in theory, but only verified through unit tests as I don't have a Telegram test environment.

Since this implementation uses the existing postMessage function for proxy handling, no additional proxy-specific tests were added.

Resolves #432

### DIFF
--- a/docs/spec/v1beta3/providers.md
+++ b/docs/spec/v1beta3/providers.md
@@ -554,8 +554,7 @@ This can be a unique identifier (`-1234567890`) for the target chat,
 a unique identifier with the topic identifier (`-1234567890:1`) for the forum chat,
 or the username (`@username`) of the target channel.
 
-This Provider type does not support the configuration of a [proxy URL](#https-proxy)
-or [certificate secret reference](#certificate-secret-reference).
+This Provider type does not support the configuration of a [certificate secret reference](#certificate-secret-reference).
 
 ###### Telegram example
 
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/PagerDuty/go-pagerduty v1.8.0
 	github.com/cdevents/sdk-go v0.4.1
 	github.com/chainguard-dev/git-urls v1.0.2
-	github.com/containrrr/shoutrrr v0.8.0
 	github.com/elazarl/goproxy v1.7.2
 	github.com/fluxcd/cli-utils v0.36.0-flux.13
 	github.com/fluxcd/notification-controller/api v1.6.0
@@ -99,7 +98,6 @@ require (
 	github.com/emicklei/go-restful/v3 v3.12.1 // indirect
 	github.com/evanphx/json-patch/v5 v5.9.11 // indirect
 	github.com/exponent-io/jsonpath v0.0.0-20210407135951-1de76d718b3f // indirect
-	github.com/fatih/color v1.16.0 // indirect
 	github.com/felixge/httpsnoop v1.0.4 // indirect
 	github.com/fluxcd/pkg/apis/acl v0.7.0 // indirect
 	github.com/fluxcd/pkg/apis/kustomize v1.10.0 // indirect
@@ -145,8 +143,6 @@ require (
 	github.com/leodido/go-urn v1.4.0 // indirect
 	github.com/liggitt/tabwriter v0.0.0-20181228230101-89fcab3d43de // indirect
 	github.com/mailru/easyjson v0.9.0 // indirect
-	github.com/mattn/go-colorable v0.1.13 // indirect
-	github.com/mattn/go-isatty v0.0.20 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
 	github.com/mitchellh/go-wordwrap v1.0.1 // indirect
 	github.com/mitchellh/mapstructure v1.5.0 // indirect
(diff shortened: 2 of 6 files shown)


matheuscscp:
I tested this change and it's not working :cry: 

I tested with both `v1.6.0` and this version. In `v1.6.0` the configuration below works:

share_920559451063054126

In this version I get this error log:

```json
{
    "level": "error",
    "ts": "2025-07-07T16:02:28.735Z",
    "logger": "event-server",
    "msg": "failed to send notification",
    "eventInvolvedObject": {
        "kind": "ResourceSetInputProvider",
        "namespace": "flux-system",
        "name": "rsip-gh-app-token-cache-test",
        "uid": "0cfaf88d-27a7-47ae-bcc5-c7870a7e0143",
        "apiVersion": "fluxcd.controlplane.io/v1",
        "resourceVersion": "614409"
    },
    "alert": {
        "name": "telegram-notifications",
        "namespace": "flux-system",
        "providerName": "telegram-bot"
    },
    "error": "request failed: request failed with status code 404, {\"ok\":false,\"error_code\":404,\"description\":\"Not Found\"}"
}
```

### Configuration

```yaml
kind: Provider
metadata:
  name: telegram-bot
  namespace: flux-system
spec:
  type: telegram
  address: https://api.telegram.org
  channel: "<redacted>"
  secretRef:
    name: telegram-bot-secret
---
apiVersion: v1
kind: Secret
metadata:
  name: telegram-bot-secret
  namespace: flux-system
type: Opaque
stringData:
  token: <redacted>
---
kind: Alert
metadata:
  name: telegram-notifications
  namespace: flux-system
spec:
  providerRef:
    name: telegram-bot
  eventSeverity: info
  eventSources:
    - kind: Kustomization
      name: '*'
      namespace: flux-system
    - kind: HelmRelease
      name: '*'
      namespace: flux-system
    - kind: ResourceSetInputProvider
      name: '*'
      namespace: flux-system
```

cappyzawa:
@matheuscscp Thanks for testing! You found an important compatibility issue.

## Problem Analysis

In v1.6.0, the Telegram notifier used shoutrrr and completely **ignored the `address` field**:
- `NewTelegram(channel, token)` - address was never passed
- shoutrrr internally generated `telegram://token@telegram` URLs
- So `address: https://api.telegram.org` had no effect

In my new implementation, I changed this to use the `address` field, but this breaks existing configurations that expect the address to be ignored.

## Question: What should be the intended behavior?

**Option 1: Maintain backward compatibility (ignore address field)**
```go
// Always ignore address and auto-generate from token
apiURL = fmt.Sprintf("https://api.telegram.org/bot%s", token)
```

**Option 2: Support standard Telegram address**
```go
// Handle empty or standard Telegram API address
if apiURL == "" || apiURL == "https://api.telegram.org" {
    apiURL = fmt.Sprintf("https://api.telegram.org/bot%s", token)  
}
// Allow custom URLs like "https://api.telegram.org/bot123456:ABC"
```

Since the address field was effectively unused in v1.6.0, I think **Option 1** might be safer for backward compatibility. What's your preference?

I can quickly fix this once we decide on the expected behavior.

cappyzawa:
Also, as a quick workaround with the current PR implementation, your configuration should work if you remove the `address` field:

```yaml
kind: Provider
metadata:
  name: telegram-bot
  namespace: flux-system
spec:
  type: telegram
  # address: https://api.telegram.org  # <- remove this line
  channel: "<redacted>"
  secretRef:
    name: telegram-bot-secret
```

This will use the default Telegram API URL generated from the token.

matheuscscp:
> There is no self-hosted Telegram offering so I think we should maintain backward compatibility and ignore the host field.

I agree with this, let's keep ignoring the field :+1: 

> Also, as a quick workaround with the current PR implementation, your configuration should work if you remove the `address` field

This actually does not work, we have an `if` in the code blocking configurations without `address`:

https://github.com/fluxcd/notification-controller/blob/3f4e962c83fa1c13924224cb80d83701fa20d477/internal/server/event_handlers.go#L392-L394

See this error log:

```json
{
    "level": "error",
    "ts": "2025-07-08T08:48:36.282Z",
    "logger": "event-server",
    "msg": "failed to dispatch notification",
    "eventInvolvedObject": {
        "kind": "ResourceSetInputProvider",
        "namespace": "flux-system",
        "name": "rsip-gh-app-token-cache-test",
        "uid": "0cfaf88d-27a7-47ae-bcc5-c7870a7e0143",
        "apiVersion": "fluxcd.controlplane.io/v1",
        "resourceVersion": "653018"
    },
    "alert": {
        "name": "telegram-notifications",
        "namespace": "flux-system",
        "providerName": "telegram-bot"
    },
    "error": "failed to initialize notifier for provider 'telegram-bot': provider has no address"
}
```

### Action we need to take in a separate PR

The Telegram provider proves that this address check is not relevant for all providers. We should remove this check from the event handler and put it inside each provider that does require an address. The Telegram provider will not be one of them.

cappyzawa:
@stefanprodan @matheuscscp Done! I've implemented your suggestion to always ignore the address field for backward compatibility.

The Telegram notifier now always generates `https://api.telegram.org/bot<TOKEN>` from the token, just like v1.6.0 did.

This should fix the reported 404 issue - existing configurations with `address: https://api.telegram.org` should work now.

matheuscscp (approved):
LGTM! 🚀

PR #1141, opened 2025-07-09
Make address field optional for providers that generate URLs internally

cappyzawa:
## Summary

This PR resolves the issue mentioned in https://github.com/fluxcd/notification-controller/pull/1140#issuecomment-3047971232, making the `address` field optional for providers that generate URLs internally.

- Removed generic address validation from `event_handlers.go` to allow address-optional providers
- Made `Provider.spec.address` and Secret `address` key optional for providers like Telegram
- Address validation is now handled by individual notifiers, and all required validations are already implemented
  - Based on my investigation, existing notifiers have sufficient URL validation, but please let me know if there are specific notifiers that need additional validation checks


matheuscscp:
Actually there's one more improvement we can do here.

With this PR, `notifier.NewFactory()` no longer has any required arguments except for `ctx`. The `url` is no longer required, so we can remove it and add a function `notifier.WithURL()`, and call this function only if `webhook != ""`.

matheuscscp:
Oops, we forgot to do something that would break notification-controller big time, we need to remove this:

https://github.com/fluxcd/notification-controller/blob/e8a909fac70913c0d2f531e397d2d671a0f89e17/internal/notifier/factory.go#L223-L225

matheuscscp:
The following constructors need to validate that the URL is not empty:

* `NewSentry`
* `NewAzureEventHub`

All the others are indeed doing this already.

matheuscscp (approved):
LGTM! 🚀 

Please squash 🙏

PR #1142, opened 2025-07-10
Fix double secret fetching in BasicAuth processing

cappyzawa:
## Summary

This PR fixes a performance regression introduced in #1139 where `spec.secretRef` was being fetched twice.

## Problem

As pointed out by @matheuscscp in https://github.com/fluxcd/notification-controller/pull/1139#issuecomment-3056595416, the implementation was fetching the secret twice:
1. Once in `extractAuthFromSecret`
2. Again in `secrets.BasicAuthFromSecret`

Since notification-controller doesn't cache secrets by default, this has a significant performance impact.

## Solution

Following Option 1 from https://github.com/fluxcd/flux2/issues/5433#issuecomment-3031563739, this PR moves the BasicAuth processing directly into `extractAuthFromSecret` using the already-fetched secret data. This aligns with the special nature of `spec.secretRef` that contains multiple authentication methods.

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -344,6 +344,13 @@ func extractAuthFromSecret(ctx context.Context, kubeClient client.Client, provid
 		options = append(options, notifier.WithHeaders(headers))
 	}
 
+	if user, ok := secret.Data["username"]; ok {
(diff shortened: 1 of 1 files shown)


matheuscscp (approved):
LGTM! 🚀

PR #1151, opened 2025-07-27
Introduce label selector for watching Secrets referenced in Receivers

matheuscscp:
xref: https://github.com/fluxcd/flux2/issues/5446


stefanprodan (approved):
LGTM 🚀

PR #1160, opened 2025-08-15
Add support for mTLS to GitHub App transport

abhijith-darshan:
`notification-controller` currently generates a TLS config for the underlying `HTTP` transport if `Provider` spec contains a `certSecretRef`.

Related to the following issues, notification-controller will also use the TLS config for mTLS GitHub App scenarios as well -

- https://github.com/fluxcd/pkg/pull/999
- https://github.com/fluxcd/source-controller/pull/1860
- https://github.com/fluxcd/image-automation-controller/pull/947

- `.spec.certSecretRef` is the default choice for extracting the `ca.crt`.
- If the `ca.crt` is present in both `.spec.certSecretRef` and `.spec.secretRef` then `.spec.certSecretRef` takes the higher precedence.
- The `ca.crt` in `.spec.secretRef` is only consider if the provider is of one of the `git` types and `certSecretRef` is not specified in the spec.

example:

```yaml
kind: Provider
....
spec:
  address: https://my.company.corp/org/repo
  certSecretRef:
    name: some-ca-secret # TLS config with custom ca is extracted from this secret
  secretRef:
    name: github-app-ca-secret # the ca.crt if present in this secret reference will be ignored
  type: github
```

TODOS:

- [ ] `ca.crt` for TLS config should be generated from `.spec.certSecretRef` and if not specified then it should attempt to generate TLS config from `.spec.secretRef` (only if the provider is one of the `git` types)
- [ ] Add relevant test cases
- [ ] Update the docs regarding the `ca.crt` preference

(chore): keep Makefile in sync with other controllers


internal/server/event_handlers.go
```diff
@@ -451,6 +447,34 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider *api
...
+	// attempt to get TLS config from secretRef if ca.crt exists in secretRef
+	if isGitProvider(provider.Spec.Type) {
+		if provider.Spec.SecretRef != nil {
+			secretRef.Name = provider.Spec.SecretRef.Name
+			tlsConfig, err = secrets.TLSConfigFromSecretRef(ctx, kubeClient, secretRef)
+			if err != nil {
+				return nil, fmt.Errorf("failed to get TLS config: %w", err)
+			}
+		}
+	}
```
matheuscscp, line 474: @cappyzawa GitRepository is the only one accepting TLS config from `.spec.secretRef` at the moment, right?

Would we like to keep it that way, or do we want to make this exception for Git commit status providers here in notification-controller too? If we follow the same logic suggested by @stealthybox for supporting the legacy TLS secret keys everywhere, maybe this change is ok with the justification of being only for Git providers to keep them consistent, like we chose for TLS.

Thoughts anybody?

@stefanprodan

internal/server/event_handlers.go
```diff
@@ -451,6 +447,34 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider *api
...
+			return nil, fmt.Errorf("failed to get TLS config: %w", err)
+		}
+		return
+	}
+	// if certSecretRef is not specified, and if the provider is a git provider then
+	// attempt to get TLS config from secretRef if ca.crt exists in secretRef
+	if isGitProvider(provider.Spec.Type) {
+		if provider.Spec.SecretRef != nil {
+			secretRef.Name = provider.Spec.SecretRef.Name
+			tlsConfig, err = secrets.TLSConfigFromSecretRef(ctx, kubeClient, secretRef)
```
matheuscscp, line 469: This secret is already being fetched inside `extractAuthFromSecret()`, let's move the `.Get()` from there to `getTLSConfigForProvider()` and pass the `*corev1.Secret` down to both `extractAuthFromSecret()` and `getTLSConfigForProvider()`.

matheuscscp (commented):
First pass :ok_hand:

internal/server/event_handlers.go
stefanprodan: I think the main use-case is for pointing the Alert Provider to the same secret you use for the GitRepository. So the code here makes sense to me.

internal/server/event_handlers.go
```diff
@@ -451,6 +454,38 @@ func createNotifier(ctx context.Context, kubeClient client.Client, provider *api
...
+// getTLSConfigForProvider - retrieves the TLS configuration from the provider's certSecretRef or secretRef.
+func getTLSConfigForProvider(ctx context.Context, providerCertSecret, providerSecret *corev1.Secret, providerType string) (tlsConfig *tls.Config, err error) {
+	// providerCertSecret takes precedence over providerSecret as it is explicitly specified for TLS configuration
+	if providerCertSecret != nil {
+		tlsConfig, err = secrets.TLSConfigFromSecret(ctx, providerCertSecret)
+		if err != nil {
+			return nil, fmt.Errorf("failed to get TLS config: %w", err)
+		}
+		if tlsConfig.RootCAs != nil {
+			log.FromContext(ctx).Info("using ca from providerCertSecret for TLS configuration")
```
matheuscscp:
```suggestion
			log.FromContext(ctx).V(1).Info("using ca from providerCertSecret for TLS configuration")
```

internal/server/event_handlers.go
abhijith-darshan: I also missed the fact that a `secretRef` can also be with no `ca.crt` or client key data.

using `secrets.AuthMethodsFromSecret` and `authMethods.HasTLS()` solves that problem

```go
	if isGitProvider(providerType) && providerSecret != nil {
		authMethods, err := secrets.AuthMethodsFromSecret(ctx, providerSecret)
		if err != nil {
			return nil, fmt.Errorf("failed to get TLS config: %w", err)
		}
		// only proceed to create TLS config if ca.crt exists in the secret
		if authMethods != nil && authMethods.HasTLS() {
                   ....
		}
	}
```

matheuscscp:
This is looking good, thanks! Please add docs and amend the commit message to match the PR title :pray:

PR #1161, opened 2025-08-17
[RFC-0010] Add default-service-account for lockdown

cappyzawa:
Part of: https://github.com/fluxcd/flux2/issues/5465


matheuscscp (approved):
LGTM! :rocket:

Issue #1144, opened 2025-07-14
Remove deprecated proxy fields from Provider API v1


cappyzawa:
# Remove deprecated proxy fields from Provider API

## Background

The deprecated proxy configuration fields need to be removed as planned. These fields were marked for deprecation in [PR #1133](https://github.com/fluxcd/notification-controller/pull/1133) when `spec.proxySecretRef` was introduced as the recommended approach.

### 1. `spec.proxy` field
- **Location**: `Provider` spec
- **Deprecated since**: v2.7 (#1133)
- **Replacement**: `spec.proxySecretRef`
- **Current warning**: "spec.proxy is deprecated, please use spec.proxySecretRef instead. Support for this field will be removed in v1."

### 2. `proxy` key in secret
- **Location**: Referenced secret data
- **Deprecated since**: v2.7 (#1133) 
- **Replacement**: `spec.proxySecretRef` with `address` key
- **Current warning**: "specifying proxy with 'proxy' key in the referenced secret is deprecated, use spec.proxySecretRef with 'address' key instead. Support for the 'proxy' key will be removed in v1."

## Timeline

- **Target**: After Flux v2.7 release (around September 2025)
- **Rationale**: Need one release cycle with deprecation warnings before removal

## Implementation Tasks

- [ ] Remove deprecated proxy fields from API and controller logic
- [ ] Optimize URL parsing to eliminate redundant parsing across notifiers
  - https://github.com/fluxcd/notification-controller/pull/1133#pullrequestreview-2966720020
- [ ] Update documentation and examples

PR #1190, opened 2025-10-06
Remove deprecated proxy fields from Provider API

cappyzawa:
ref: #1144 .

Note: The proxy URL parsing optimization mentioned in #1144 will be addressed in a separate follow-up PR.


matheuscscp:
> Adding SOCKS5 support would be a feature addition, not suitable for a patch release.

Small correction: Making the validation stricter isn't suitable for a patch release (unless we're fixing a CVE, which was not the case), so we decided to revert the validation PR in `fluxcd/pkg` altogether. Support for SOCKS5 already existed (removing it was the regression).

PR #1077, opened 2025-03-19
Update docs to include Telegram forum configuration

v1km4n:
Current documentation does not mention that Telegram provider supports Telegram forum chats' topics, which it does, since [the shoutrrr lib does](https://github.com/containrrr/shoutrrr/pull/362/files). Therefore, no changes are needed in the FluxCD Notification Controller in order to send Telegram alerts to said topics.

This PR contains documentation changes that describe how you can set Telegram provider up to work with Telegram forum chats.

Closes [#1063](https://github.com/fluxcd/notification-controller/issues/1063)

PR #1199, opened 2025-10-18
Fix support for telegram message_thread_id

### LINKED ISSUE #1063: Request to Support message_thread_id Parameter for Telegram Provider
**Summary**
The current Telegram provider in the notification-controller does not support sending alerts to Telegram forums (chats with topics).

This feature can be implemented by extending the Telegram provider to accept the message_thread_id parameter in its notification configuration.

**Suggested Implementation**
The Telegram Bot API supports the message_thread_id parameter for the sendMessage method (as detailed in the [Telegram Bot API documentation](https://core.telegram.org/bots/api#sendmessage)).

**Current behavior**: The Telegram provider sends messages to the default chat without the option to specify a topic. As a result there is currently no way to utilize Telegram chats that use topics feature.

**Desired behavior**: Add support for the message_thread_id parameter to the configuration of the Telegram provider so that notifications can be sent to specific threads in a Telegram forum. Provider manifest will the look like so:

```
kind: Provider
metadata:
  name: telegram-alerts
spec:
  type: telegram
  channel: "*chat_id*"
  message_thread_id: "*message_thread_id*"
  secretRef:
    name: telegram-bot-token
```

Kasama:
Hello!

Since #1140, the notification controller stopped using [shoutrrr](https://github.com/containrrr/shoutrrr) to send telegram messages.

This change has caused #1077 to be no longer valid and #1063 to resurface.

This PR aims to fix #1063 again by implementing the same semantics that are written in the [documententation](https://fluxcd.io/flux/components/notification/providers/#telegram).

### DIFF
--- a/internal/notifier/telegram.go
+++ b/internal/notifier/telegram.go
@@ -25,9 +25,10 @@ type Telegram struct {
 // TelegramPayload represents the payload sent to Telegram Bot API
 // Reference: https://core.telegram.org/bots/api#sendmessage
 type TelegramPayload struct {
-	ChatID    string `json:"chat_id"`    // Unique identifier for the target chat
-	Text      string `json:"text"`       // Text of the message to be sent
-	ParseMode string `json:"parse_mode"` // Mode for parsing entities in the message text
+	ChatID          string `json:"chat_id"`                     // Unique identifier for the target chat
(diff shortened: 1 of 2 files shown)


matheuscscp (commented):
Hello @Kasama :wave: Thanks very much for finding and fixing this :hugs:

PR #1238, opened 2026-02-14
Update pkg deps to flux/v2.8.x versions

matheuscscp:
Fixing my mistake in the review from earlier, `pkg/ssa` was branched for Flux 2.8 from `v0.67.0`.

Dipti is fixing this in KC/HC.


stefanprodan (approved):

matheuscscp:
Going to let the new workflow do this

PR #1287, opened 2026-04-10
Remove GCR audience reconstruction

matheuscscp:
Follow up for #1279.

The `audience` must also be present on the secret. (Just like `email`, it will become required in Flux 2.9.)


stefanprodan (approved):
LGTM

PR #1290, opened 2026-04-10
Fix generic providers not forwarding commit status events

### LINKED ISSUE #1289: v1.8.0 no longer sends ReconciliationSucceeded events to the Generic Webhook Provider
First off I'll say I'm not sure if this is a bug or if we were using the notification-controller incorrectly

From version v1.8.0 of the notification-controller, a change was made in [#1231](https://github.com/fluxcd/notification-controller/pull/1231/changes#diff-7e6eaf79d30a5a94c022c0ba49c3d4c345a74e7fa64700397ec109e23ba59583R294) to skip sending events if the event contains `kustomize.toolkit.fluxcd.io/commit_status` in the metadata.

Until 1.8.0 these events would be sent to the Generic Provider. I do see that previously the skip commit metadata updates were filtered out in each notifier rather than directly in the notifiers themselves.

Somewhat related is that the logs appear to show that the event was sent to the provider. Whereas other events which aren't sent have the message "discarding event, ...". 
```json
{"level":"info","ts":"2026-04-10T10:29:58.421Z","logger":"event-server","msg":"dispatching event","eventInvolvedObject":{"kind":"Kustomization","namespace":"flux-system","name":"my-kustomization","uid":"c0664df8-21db-453c-a628-31d09cb35821","apiVersion":"kustomize.toolkit.fluxcd.io/v1","resourceVersion":"244318194"},"message":"Reconciliation finished in 449.617964ms, next run in 24h0m0s"}
```

matheuscscp:
Fixes: #1289

### DIFF
--- a/internal/server/event_handlers.go
+++ b/internal/server/event_handlers.go
@@ -316,9 +316,10 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
 		return nil, droppedProviders{}, nil
 	}
 
-	// Skip if the event has commit status update metadata but the provider is not a git provider.
-	// Git providers (github, gitlab, etc.) are the ones that set commit statuses.
-	if !isCommitStatusProvider(provider.Spec.Type) && isCommitStatusUpdate(event) {
+	// Skip if the event has commit status update metadata but the provider is not a git provider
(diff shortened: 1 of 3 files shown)


stefanprodan (approved):
LGTM

PR #1292, opened 2026-04-10
Add Opsgenie alias field for alert deduplication

### LINKED ISSUE #460: Opsgenie Alias for deduplication of alerts

cairon-ab:
## Summary

Set the `alias` field on Opsgenie alerts using a SHA-256 hash of the involved object's Kind, Namespace, Name, and event Reason. This allows Opsgenie to deduplicate repeated alerts for the same source instead of creating new pages for each firing.

## Changes

- Added `Alias` field to `OpsgenieAlert` struct with `json:"alias,omitempty"`
- Added `generateOpsgenieAlias()` function that creates a deterministic SHA-256 hash from the event's `InvolvedObject.Kind`, `InvolvedObject.Namespace`, `InvolvedObject.Name`, and `Reason`
- Wired the alias into `Post()` so every alert carries a stable alias
- Added comprehensive tests: `TestOpsgenie_PostAlias` (verifies alias in HTTP payload across multiple scenarios) and `TestGenerateOpsgenieAlias` (determinism, length, uniqueness)

## Why

Currently, Opsgenie creates a new alert for every notification event, even when it's for the same object and reason. The [Opsgenie Alert API](https://docs.opsgenie.com/docs/alert-api) supports an `alias` field that acts as a deduplication key — alerts with the same alias are grouped as a single incident.

By hashing `Kind/Namespace/Name/Reason`, we ensure:
- **Determinism**: Same event source always produces the same alias
- **Deduplication**: Repeated firings update the existing alert instead of paging again
- **Differentiation**: Different reasons (e.g. `ReconciliationFailed` vs `HealthCheckFailed`) create separate alerts
- **Size safety**: The 64-char hex hash stays well within Opsgenie's 512-char alias limit

Fixes #460

### DIFF
--- a/internal/notifier/opsgenie.go
+++ b/internal/notifier/opsgenie.go
@@ -29,10 +29,11 @@ import (
 )
 
 type Opsgenie struct {
-	URL       string
-	ProxyURL  string
-	TLSConfig *tls.Config
-	ApiKey    string
+	URL         string
+	ProxyURL    string
+	TLSConfig   *tls.Config
+	ApiKey      string
+	ProviderUID string
 }
 
 type OpsgenieAlert struct {
@@ -42,7 +43,7 @@ type OpsgenieAlert struct {
 	Details     map[string]string `json:"details"`
 }
 
-func NewOpsgenie(hookURL string, proxyURL string, tlsConfig *tls.Config, token string) (*Opsgenie, error) {
+func NewOpsgenie(hookURL string, proxyURL string, tlsConfig *tls.Config, token string, providerUID string) (*Opsgenie, error) {
 	_, err := url.ParseRequestURI(hookURL)
 	if err != nil {
 		return nil, fmt.Errorf("invalid Opsgenie hook URL %s: '%w'", hookURL, err)
@@ -53,10 +54,11 @@ func NewOpsgenie(hookURL string, proxyURL string, tlsConfig *tls.Config, token s
 	}
 
 	return &Opsgenie{
-		URL:       hookURL,
-		ProxyURL:  proxyURL,
-		ApiKey:    token,
-		TLSConfig: tlsConfig,
+		URL:         hookURL,
+		ProxyURL:    proxyURL,
+		ApiKey:      token,
+		TLSConfig:   tlsConfig,
+		ProviderUID: providerUID,
 	}, nil
 }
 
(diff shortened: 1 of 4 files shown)


internal/notifier/opsgenie.go
```diff
@@ -91,3 +100,20 @@ func (s *Opsgenie) Post(ctx context.Context, event eventv1.Event) error {
...
+// deduplicate alerts — alerts with the same alias are treated as the same
+// incident instead of creating new pages. The alias is a SHA-256 hash
+// (truncated to 64 chars) to stay within Opsgenie's 512-char alias limit
+// while remaining collision-resistant.
+func generateOpsgenieAlias(event eventv1.Event) string {
+	key := fmt.Sprintf("%s/%s/%s/%s",
+		event.InvolvedObject.Kind,
+		event.InvolvedObject.Namespace,
+		event.InvolvedObject.Name,
+		event.Reason,
```
stefanprodan, line 115: This will cause alerts from multiple clusters to aggregate under the same incident since all clusters have `Kustomization/flux-system/flux-system`, which is a major breaking change. Adding the Alert Provider UID to the checksum would ensure each cluster gets a dedicated incident.

cairon-ab: Good catch — in a multi-cluster setup `Kustomization/flux-system/flux-system` would hash identically across clusters, collapsing separate incidents into one.

`ProviderUID` is already available in `notifierOptions` but `opsgenieNotifierFunc` doesn't pass it through to `NewOpsgenie`. I'll thread it into the `Opsgenie` struct and prepend it to the alias hash input so each Provider (and therefore each cluster) gets a unique alias. Will push the fix shortly.

matheuscscp: this looks like AI talking

cairon-ab: Fair point — I did lean on AI for drafting that reply. The fix is in f322c46b though: threads ProviderUID into the Opsgenie struct and prepends it to the alias hash, so each cluster's Provider produces distinct aliases. Added tests for the multi-cluster case. Happy to adjust if the approach doesn't look right.

stefanprodan (approved):
LGTM

PR #1306, opened 2026-05-19
Introduce `generic-oidc` Receiver

### LINKED ISSUE #1305: Feature Request: Support OIDC for generic receiver (generic, secure, secret-less Receiver)
Today we have the `generic` Receiver, which takes a secret only for deterministic generation of an HTTP path. In this case, the secret is not really secret, as the HTTP path is public and anybody/anything with network access to notification-controller can trigger the Receiver. This is fairly insecure.

We have also `generic-hmac`, which makes the secret actually a secret. Only callers that know this shared secret can trigger the Receiver. This is what I'm using in my GitOps pipelines at the moment.

We could introduce `generic-oidc`. All my GitOps pipelines have support for issuing OIDC tokens (like most CI systems at this point), so I'd like to switch to this solution, which is fully secret-less and secure.

### Proposal

```yaml
spec:
  type: generic-oidc
  oidcProviders:
    - issuerURL: https://token.actions.githubusercontent.com
      audience: notification-controller
      variables:
        - name: owner
          expression: "claims.repository_owner"
        - name: repo
          expression: "claims.repository"
        - name: allowedRepos
          expression: "['my-org/repo-a', 'my-org/repo-b']"
      validations:
        - expression: "vars.owner == 'my-org'"
          message: "wrong org"
        - expression: "vars.repo in vars.allowedRepos"
          message: "repo not in allowlist"
        - expression: "claims.ref == 'refs/heads/main' && claims.environment == 'production'"
          message: "must be main branch + production environment"
```

matheuscscp:
Closes: #1305


stefanprodan (approved):
LGTM

Thanks @matheuscscp 🥇

PR #1310, opened 2026-05-21
Make Receiver `secretRef` optional for `generic-oidc`

matheuscscp:
Follow-up for #1306

Today, the `token` inside the secret must be present but can be empty. I was planning to set an empty token in the secret to use in my clusters for `generic-oidc`, but then I realized that we can just make the API simpler and not require the `secretRef` to be set only for `generic-oidc`.

### DIFF
--- a/docs/spec/v1/receivers.md
+++ b/docs/spec/v1/receivers.md
@@ -274,8 +274,6 @@ metadata:
   namespace: default
 spec:
   type: generic-oidc
-  secretRef:
-    name: webhook-token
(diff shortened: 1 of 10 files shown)


docs/spec/v1/receivers.md
stefanprodan, line 235: Remove the part about the Secret reference

matheuscscp: It's still used, but optional. Defaults to `""`.

PR #1311, opened 2026-05-21
Reject `secretRef` for `generic-oidc` Receiver

matheuscscp:
Follow-up for #1306 

Supersedes #1310 

Make it fully secret-less!


stefanprodan (approved):
LGTM 🚀

PR #1315, opened 2026-06-01
feat: Allow receiver to notify resources in any namespace

### LINKED ISSUE #1217: Receiver cannot reconcile ImageRepository resources in other namespaces
Hi, I have a question about Receiver.

## Summary
Is it currently supported for a Receiver in one namespace to trigger reconciliation of ImageRepository resources that exist in other namespaces using resourceFilter and matchLabels? Or is the Receiver limited to searching only within its own namespace?

## Background
In our cluster, we would like to trigger the reconciliation of multiple ImageRepository resources located in different namespaces from a single Receiver endpoint.

Concretely, the Receiver is placed in the `flux-system` namespace, while ImageRepository resources are expected to be created by developers across various namespaces.

Is it possible for a single Receiver endpoint to trigger the reconciliation of ImageRepository resources that exist in different namespaces by using resourceFilter?

## Configuration
The Receiver is placed under `flux-system` as follows.
We specify the name in resourceFilter so that only specific ImageRepository objects are reconciled.

```yaml
apiVersion: notification.toolkit.fluxcd.io/v1
kind: Receiver
metadata:
  name: generic-receiver
  namespace: flux-system
spec:
  type: generic
  secretRef:
    name: webhook-token
  resources:
    - apiVersion: image.toolkit.fluxcd.io/v1
      kind: ImageRepository
      name: "*"
      matchLabels:
        registry-type: gcp
  resourceFilter: "req.name.contains(res.metadata.name)"
```

We have multiple ImageRepository resources, each placed in its own namespace. Below are two sample resources (actual clusters contain more). Note that each resource is in a separate namespace.

```yaml
apiVersion: image.toolkit.fluxcd.io/v1
kind: ImageRepository
metadata:
  name: sandbox-app
  namespace: sandbox-app
  labels:
    registry-type: gcp
spec:
  provider: gcp
  image: asia-northeast1-docker.pkg.dev/sandbox-project/sandbox-app/app
  interval: 30m
---
apiVersion: image.toolkit.fluxcd.io/v1
kind: ImageRepository
metadata:
  name: sandbox-app-2
  namespace: sandbox-app-2
  labels:
    registry-type: gcp
spec:
  provider: gcp
  image: asia-northeast1-docker.pkg.dev/sandbox-project/sandbox-app-2/app
  interval: 30m
```

The incoming request looks like this (we plan to send this from our monorepo CI system):
```
$ curl -XPOST -H "Content-Type: application/json" \
  -d '{"name": "sandbox-app"}' \
  https://webhook-receiver.example.com/hook/hogehoge
```

## Expected Behavior
Both `sandbox-app` and `sandbox-app-2` ImageRepository resources should be reconciled, even though they exist in different namespaces.

## Actual Behavior / Error
```log
notification-controller-5df7f9bc66-m4b9l manager {"level":"info","ts":"2025-11-27T16:43:26.226Z","logger":"receiver-server","msg":"handling request: a8a902f2c58facb85b4997acde74ac3aa0a27764b068c9667d1xxxxxxxxxx"}
notification-controller-5df7f9bc66-m4b9l manager {"level":"error","ts":"2025-11-27T16:43:26.337Z","logger":"receiver-server","msg":"error annotating resources","reconciler kind":"Receiver","name":"generic-receiver","namespace":"flux-system","error":"no \"ImageRepository\" resources found with matching labels map[\"registry-type\":\"gcp\"]' in \"flux-system\" namespace"}
```

It appears that the controller is only searching the `flux-system` namespace, even though `--watch-all-namespaces=true` is enabled.

And, I tried adding the `namespace: "*"`.
```yaml
...
  resources:
    - apiVersion: image.toolkit.fluxcd.io/v1
      kind: ImageRepository
      name: "*"
      namespace: "*"
      matchLabels:
        registry-type: gcp
  resourceFilter: "req.name.contains(res.metadata.name)"
```

However, the following error occurred:
```
notification-controller-8454968d77-m7lsx manager {"level":"error","ts":"2025-11-27T17:01:45.937Z","logger":"receiver-server","msg":"error annotating resources","reconciler kind":"Receiver","name":"generic-receiver","namespace":"flux-system","error":"no \"ImageRepository\" resources found with matching labels map[\"registry-type\":\"gcp\"]' in \"*\" namespace"}
```

## Version and Controller Settings
notification-controller version: v1.7.4

Controller arguments include:
```yaml
- --watch-all-namespaces=true
- --no-cross-namespace-refs=false
```

## Related

- https://github.com/fluxcd/notification-controller/issues/491
- https://github.com/fluxcd/notification-controller/issues/461

project0:
This change introduces '*' as allowed namespace value for resources in the reciver. It allows notifying resources across many namespaces, this is especially very powerful in combination with label filter to establish a generic pattern without adding huge complexity with many webhooks.

fixes #1217

### DIFF
--- a/api/v1/reference_types.go
+++ b/api/v1/reference_types.go
@@ -36,6 +36,8 @@ type CrossNamespaceObjectReference struct {
 	Name string `json:"name"`
 
 	// Namespace of the referent
+	// If multiple resources across all namespaces are targeted `*` may be set,
+	// which requires the name to also be set to `*`.
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=253
 	// +kubebuilder:validation:Optional
--- a/config/crd/bases/notification.toolkit.fluxcd.io_alerts.yaml
+++ b/config/crd/bases/notification.toolkit.fluxcd.io_alerts.yaml
@@ -116,7 +116,10 @@ spec:
                       minLength: 1
                       type: string
                     namespace:
-                      description: Namespace of the referent
+                      description: |-
+                        Namespace of the referent
+                        If multiple resources across all namespaces are targeted `*` may be set,
+                        which requires the name to also be set to `*`.
                       maxLength: 253
                       minLength: 1
                       type: string
@@ -335,7 +338,10 @@ spec:
                       minLength: 1
                       type: string
                     namespace:
-                      description: Namespace of the referent
+                      description: |-
+                        Namespace of the referent
+                        If multiple resources across all namespaces are targeted `*` may be set,
+                        which requires the name to also be set to `*`.
                       maxLength: 253
                       minLength: 1
                       type: string
--- a/config/crd/bases/notification.toolkit.fluxcd.io_receivers.yaml
+++ b/config/crd/bases/notification.toolkit.fluxcd.io_receivers.yaml
@@ -216,7 +216,10 @@ spec:
                       minLength: 1
                       type: string
                     namespace:
-                      description: Namespace of the referent
+                      description: |-
+                        Namespace of the referent
+                        If multiple resources across all namespaces are targeted `*` may be set,
+                        which requires the name to also be set to `*`.
                       maxLength: 253
                       minLength: 1
                       type: string
@@ -457,7 +460,10 @@ spec:
                       minLength: 1
                       type: string
                     namespace:
-                      description: Namespace of the referent
+                      description: |-
+                        Namespace of the referent
+                        If multiple resources across all namespaces are targeted `*` may be set,
+                        which requires the name to also be set to `*`.
                       maxLength: 253
                       minLength: 1
                       type: string
--- a/docs/api/v1/notification.md
+++ b/docs/api/v1/notification.md
@@ -273,7 +273,9 @@ string
 </td>
 <td>
 <em>(Optional)</em>
-<p>Namespace of the referent</p>
+<p>Namespace of the referent
+If multiple resources across all namespaces are targeted <code>*</code> may be set,
+which requires the name to also be set to <code>*</code>.</p>
 </td>
 </tr>
 <tr>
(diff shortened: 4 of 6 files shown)


matheuscscp:
I don't think allowing cross-tenant triggering is the right solution. If any, a solution I'd consider would be a new API called `ClusterReceiver` that is cluster-scoped and therefore targeted to cluster admins, not tenants.

What I do in my clusters is to use Flux Operator's ResourceSet to easily provision a Receiver for each namespace:

https://fluxoperator.dev/docs/crd/resourceset/

Combined with this upcoming feature in Flux 2.9 you won't even  need secrets anymore  for a secure Receiver:

https://github.com/fluxcd/notification-controller/issues/1305

project0:
The problem with creating receiver per tenant/namespace, you need to wire the new webhook endpoints also on the other side. This means a lot of extra effort to either automate or manual setup :-(

project0:
> I don't think allowing cross-tenant triggering is the right solution.

I do not see a real problem with this approach. Fact is, you can already cross-namespace as of today when you know the namespace name. So your concerns is not valid in this case as this is already happening. 

 This just extends capability to be more flexible with label filter across many namespaces

matheuscscp:
> you need to wire the new webhook endpoints also on the other side. This means a lot of extra effort to either automate or manual setup :-( 

Can you pls expand on this? We just implemented `flux trigger receiver` in the CLI that should simplify this process:

https://github.com/fluxcd/flux2/pull/5908

project0:
> Can you pls expand on this? We just implemented flux trigger receiver in the CLI that should simplify this process:

I cannot follow your thoughts about the cli command, it serves actually different purpose in my opinion and applicable for any case. I dont see why i would need to execute github actions for something that i already retrieve natively as a event from the sub-system i need. How am i supposed to trigger the receiver from outside the cluster (github actions) when there is no public kubernetes controller? 

We have a github org-wide webhook per cluster (And ECR Push events webhook to notify image updates for all registries). I just want to route the webhooks to its destination within the multi-tenant namespace so i dont have to create lots of webhooks in github and ECR (either via complex script/dynamic setup whatever or manual effort). 
Managing webhooks at several places is not a trivial task and the less i have to maintain the better.

project0:
I am not saying i like the current option and also would rather like to see better native option with clusterReceiver and proper RBAC, but its what we have currently.

PR #1317, opened 2026-06-04
Fix error events being dropped by commit status providers

matheuscscp:


stefanprodan (approved):
LGTM

Thanks @matheuscscp

PR #1334, opened 2026-06-23
feat(api): add CRD categories for Alert and Provider

amarkdotdev:
## Summary

Part of fluxcd/flux2#5947.

Adds `categories: [all, fluxcd, fluxcd-notifications]` to the Alert and Provider storage-version CRDs so they can be listed with `kubectl get fluxcd` and `kubectl get fluxcd-notifications`.

## Testing

- `make manifests`
- Verified generated CRD YAML contains the expected categories


matheuscscp:
Duplicate of: #1333

PR #1353, opened 2026-07-26
Add incident.io alert provider

### LINKED ISSUE #1124: Incidentio notification provider
Hi Flux team! I would like to know if there is any plan to add IncidentIO for notification-controller? We moved away from Pagerduty and so was considering if that's feasible. I could always use slack or datadog and then hook them up to our alerting service but was wondering if it's on the map. Happy to contribute as well!

xchose:
Fixes #1124.

Adds the `incident.io` Provider type posting alert events to an [incident.io HTTP alert source](https://docs.incident.io/api-reference/alert-events-v2) (bearer-token auth, `202 Accepted` response).

### Design decisions

1. **Severity mapping**: `error` events are sent with status `firing`, all other severities with status `resolved` — the same semantics as the PagerDuty provider, enabling automatic alert resolution in incident.io.
2. **Deduplication key**: the involved object's UID — same approach as the PagerDuty provider. A name-based key (`<kind>/<namespace>/<name>`) would collide across clusters reporting to the same alert source (every cluster fires `Kustomization/flux-system/flux-system`), letting an info event from one cluster resolve another cluster's alert. Changed after review feedback.
3. `Progressing` events are skipped, matching the PagerDuty provider behavior.
4. Docs note that `eventSeverity: error` on an Alert disables auto-resolution (info events must flow for recovery to be delivered).

### Testing

- Unit tests covering payload mapping, bearer auth header, 202 handling, firing/resolved mapping, Progressing skip, and constructor validation; fuzz harness consistent with the other providers.
- Tested end-to-end on a kind cluster against a real incident.io workspace: alert fires on a broken GitRepository branch, auto-resolves on recovery via the deduplication key, separate objects produce separate alerts, and an invalid token surfaces as a 401 error in the controller logs.
- Retested after the dedup-key change: captured the raw payload with an in-cluster echo server — firing and recovery events carry the same object UID as `deduplication_key`, and a second object gets its own UID key (no sharing). Screenshots in the review thread show the alert detail with the UID key.

Alert fired for the broken GitRepository branch:

incident.io alert list showing the alert in Firing state

Automatically resolved after the recovery event (same deduplication key):

incident.io alert list showing both alerts Resolved

Alert detail with event message, deduplication key and firing→resolved timeline:

incident.io alert detail with deduplication key and resolved status

Step-by-step testing procedure on a kind cluster:
[INCIDENTIO_E2E_REPORT.md](https://github.com/user-attachments/files/30391795/INCIDENTIO_E2E_REPORT.md)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


internal/notifier/incidentio.go
```diff
@@ -0,0 +1,126 @@
...
+	metadata["severity"] = event.Severity
+	metadata["reason"] = event.Reason
+	if event.ReportingController != "" {
+		metadata["reportingController"] = event.ReportingController
+	}
+
+	payload := IncidentioPayload{
+		Title:            fmt.Sprintf("%s: %s", objName, event.Reason),
+		Status:           status,
+		DeduplicationKey: fmt.Sprintf("%s/%s/%s", kind, obj.Namespace, obj.Name),
```
stefanprodan: All clusters in your fleet will fire `Kustomization/flux-system/flux-system`, and same for all your apps and infra alerts. Why would you want an info event coming from a dev cluster to close a prod incident?

xchose: Thanks for the review @stefanprodan! Good catch — the deduplication key is now the involved object's UID instead of `<kind>/<namespace>/<name>`, following the same approach as the PagerDuty provider, so identically named objects on different clusters can never share a key and an event from one cluster can no longer resolve another cluster's alert.

Re-ran the e2e test against a real incident.io workspace with the updated image: the firing payload carries the object UID as `deduplication_key`, and the recovery event with the same UID auto-resolves the alert.
1-alerts-list-firing
2-alert-detail-firing-uid-dedup-key
3-alert-detail-resolved


Ready for another review when you have a moment 🙏

stefanprodan: Please update the PR description

xchose: Done — PR description updated to reflect the UID-based deduplication key, and the attached e2e report now includes the UID retest.

stefanprodan (approved):
LGTM

Thanks @xchose 🏅

matheuscscp:
Changed provider type to `incident.io` in #1360

PR #1358, opened 2026-08-02
test: cover 2xx webhook responses including 204

### LINKED ISSUE #441: 204 status in generic webhook is treated as error
When a webhook configured through the `generic` or `generic-hmac` Provider returns an HTTP status of 204 an error is logged:

```
{"level":"error","ts":"2022-11-16T14:28:39.167Z","logger":"event-server","msg":"failed to send notification","reconciler kind":"HelmRelease","name":"metallb","namespace":"metallb","error":"postMessage failed: request failed with status code 204, "}
```

I think we should treat any 2xx status as error.

zanarellidev:
## Summary
- The default `postMessage` response validator already accepts any HTTP 2xx status (`200 <= s < 300`), including `204 No Content`.
- The in-code comment still said only 200/201/202, which matches the stale claim in #441.
- Add regression tests for 200/201/202/204 success and 400 failure, and update the comment.

Fixes #441

## Test plan
- [x] `go test -count=1 -run Test_postMessage_accepts2xxStatusCodes ./internal/notifier/`
- [x] Existing `Test_postMessage` still passes

### DIFF
--- a/internal/notifier/client.go
+++ b/internal/notifier/client.go
@@ -43,7 +43,7 @@ type postOption func(*postOptions)
 
 func postMessage(ctx context.Context, address string, payload any, opts ...postOption) error {
 	options := &postOptions{
-		// Default validateResponse function verifies that the response status code is 200, 202 or 201.
+		// Default validateResponse accepts any 2xx status (including 204 No Content).
 		responseValidator: func(resp *http.Response) error {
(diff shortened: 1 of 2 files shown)


stefanprodan (approved):
LGTM

Thanks @zanarellidev

PR #1360, opened 2026-08-05
Change incident.io provider type

matheuscscp:
This updates the incident.io Provider type value from `incidentio` to `incident.io` so the enum looks better in YAML.

Follow-up for  #1353


stefanprodan (approved):
LGTM

PR #1369, opened 2026-08-11
Sort event metadata in notifications

artur-chopikian:
Notification fields are built by ranging over `event.Metadata`. Go randomises map iteration, so the same event produces a different field order on every send:

```
helmrelease/purple-backend.purple    cluster / env / revision / image-tag
helmrelease/purple-webcli.purple     image-tag / cluster / env / revision
```

Same four keys, two orders, seconds apart. The value being looked for moves between messages.

`teams.go` already sorts its Adaptive Card facts with `slices.SortFunc`, so the guarantee exists — it was not applied elsewhere, including to the deprecated connector payload in the same file.

Applied to the notifiers whose output is read in order: slack, discord, rocket, teams, google_chat, zoom, webex, grafana, lark, matrix, telegram. The iterate-sorted-keys form is used throughout because five of them concatenate strings and cannot be sorted afterwards.

Not applied to incidentio and sentry, where metadata goes into a map, nor to datadog and otel, where tags and attributes are sets.

`TestSlack_PostFieldsAreSorted` posts one event ten times and asserts the order each time. It fails on main. `make tidy fmt vet` and `make test` pass, working tree clean.

### DIFF
--- a/internal/notifier/discord.go
+++ b/internal/notifier/discord.go
@@ -19,8 +19,10 @@ package notifier
 import (
 	"context"
 	"fmt"
+	"maps"
 	"net/url"
 	"path"
+	"slices"
 	"strings"
 
 	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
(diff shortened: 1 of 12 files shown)


artur-chopikian:
Hello! @stefanprodan, can you check this one? Any comments?

stefanprodan (approved):
LGTM

Thanks @artur-chopikian

PR #1372, opened 2026-08-16
Add Mastodon alert provider

### LINKED ISSUE #447: Add mastodon as a notification channel
[Mastodon](https://joinmastodon.org) is free and open-source software for running self-hosted social networking services. There is a Go SDK of the Mastodon API[^1], so we can add Mastodon as a new notification channel.

Similar work has been done in the GoReleaser project[^2].

[^1]: https://github.com/mattn/go-mastodon
[^2]: https://github.com/goreleaser/goreleaser/pull/3567

/cc @dentrax

xchose:
Fixes #447

Adds a `mastodon` Provider type for posting Flux events as statuses on a Mastodon account, via a plain HTTP `POST` to the [`/api/v1/statuses`](https://docs.joinmastodon.org/methods/statuses/#create) endpoint. No SDK dependency; like every other notifier it only needs the shared `postMessage` client.

### Implementation

- The `address` is the server root URL (e.g. `https://mastodon.social`); the `/api/v1/statuses` path is appended automatically (and preserved if already present).
- Auth: OAuth access token with the `write:statuses` scope (Secret key `token`), sent as a bearer token.
- Status text: severity emoji (`💫`/`🚨`, same convention as the Telegram provider), involved object `kind/name.namespace`, event message, and event metadata as key-value lines. Truncated to 500 characters, the default Mastodon server limit.
- The request payload carries only `status`. Visibility follows the account's default posting visibility (Preferences → Posting defaults); "Quiet public" or "Followers only" is recommended for alert accounts and documented in the spec.
- The `Idempotency-Key` header carries the event key the event server already uses for rate limiting. The key computation moved from `eventKeyFunc` to `notifier.EventKey`, is computed once in `eventMiddleware` and passed to notifiers through the context (the notifier cannot recompute it, since the token metadata is stripped before dispatch). Mastodon keeps the key for one hour, so retried requests and identical events within that window do not create additional statuses.
- Proxy and TLS configuration from the Provider spec are respected.

Sample rendered payload:

```json
{
  "status": "💫 gitrepository/podinfo.flux-system\nstored artifact for commit 'master@sha1:3e0ff8a...'\n\nrevision: master@sha1:3e0ff8a..."
}
```

### Testing

- `internal/notifier/mastodon_test.go`: endpoint path, bearer auth header, `Idempotency-Key` taken from the context and omitted without it, payload shape without `visibility`, severity emoji, 500-char truncation, sorted metadata, path preservation and constructor validation against an `httptest` server.
- `internal/notifier/eventkey_test.go`: key stability and context round-trip.
- `internal/controller/provider_controller_test.go`: the `mastodon` type is accepted by the API server.
- Full envtest suite passes; `make tidy fmt vet generate manifests api-docs` leave a clean tree.

### Manual e2e testing

kind cluster, controller image and CRD from this branch, real mastodon.social account. Secret: `address: https://mastodon.social`, `token` with `write:statuses`. Verified against an in-cluster echo server and then against mastodon.social:

- payload is `{"status": ...}` only, `Idempotency-Key` equals the rate-limiter digest recomputed by hand and is stable across identical events;
- info, error and recovery events from a podinfo `GitRepository` posted three statuses without errors;
- statuses were `public` before changing the account setting and `unlisted` after switching it to "Quiet public";
- a repeated identical event past a 10s rate limit created no duplicate status.

mastodon-info

mastodon-error

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/spec/v1beta3/providers.md
+++ b/docs/spec/v1beta3/providers.md
@@ -97,6 +97,7 @@ The supported alerting providers are:
 | [Google Pub/Sub](#google-pubsub)                        | `googlepubsub`   |
 | [Grafana](#grafana)                                     | `grafana`        |
 | [incident.io](#incidentio)                              | `incident.io`    |
+| [Mastodon](#mastodon)                                   | `mastodon`       |
 | [Lark](#lark)                                           | `lark`           |
 | [Matrix](#matrix)                                       | `matrix`         |
 | [Microsoft Teams](#microsoft-teams)                     | `msteams`        |
@@ -1417,6 +1418,63 @@ metadata:
 stringData:
   address: https://api.incident.io/v2/alert_events/http/<alert_source_config_id>
   token: <incident.io API token>
+```
+
+##### Mastodon
+
+When `.spec.type` is set to `mastodon`, the controller will publish an
+[Event](events.md#event-structure) as a
+[status](https://docs.joinmastodon.org/methods/statuses/#create) on the
+Mastodon account owning the referenced access token.
+
+The [Address](#address) is the Mastodon server root URL, e.g.
+`https://mastodon.social` (the `/api/v1/statuses` path is appended
+automatically). An optional `visibility` query parameter (`public`,
+`unlisted` or `private`) overrides the app's default status visibility,
+e.g. `https://mastodon.social?visibility=unlisted`.
+
+The status text contains the involved object, the event message and the event
+metadata as key-value lines. Statuses longer than 500 characters (the default
+Mastodon server limit) are truncated. An `Idempotency-Key` header derived from
+the event is sent to prevent duplicate statuses on retried requests.
+
+The access token must be provided in the `token` key of the referenced Secret,
+it is sent as a bearer token in the `Authorization` header of the POST request.
+The token can be generated in the Mastodon web interface under
+`Preferences → Development → New application` and requires the
+`write:statuses` scope.
+
+This Provider type does support the configuration of a [proxy URL](#https-proxy)
+and [certificate secret reference](#certificate-secret-reference).
+
+###### Mastodon example
+
+To configure a Provider for Mastodon, create an application with the
+`write:statuses` scope in the Mastodon web interface to obtain the access
+token, then create a Secret with [the `address`](#address-example) set to the
+server root URL, [the `token`](#token-example) set to the access token, and a
+`mastodon` Provider with a [Secret reference](#secret-reference).
+
+```yaml
+---
+apiVersion: notification.toolkit.fluxcd.io/v1beta3
+kind: Provider
+metadata:
+  name: mastodon
+  namespace: default
+spec:
+  type: mastodon
+  secretRef:
+    name: mastodon-app
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mastodon-app
+  namespace: default
+stringData:
+  address: https://mastodon.social
+  token: <Mastodon access token>
 ```
 
 
--- a/internal/notifier/mastodon.go
+++ b/internal/notifier/mastodon.go
@@ -1,0 +1,144 @@
+/*
+Copyright 2026 The Flux authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package notifier
+
+import (
+	"context"
+	"crypto/sha256"
+	"crypto/tls"
+	"errors"
+	"fmt"
+	"maps"
+	"net/url"
+	"slices"
+	"strings"
+
+	eventv1 "github.com/fluxcd/pkg/apis/event/v1beta1"
+	"github.com/hashicorp/go-retryablehttp"
+)
+
+// mastodonStatusesPath is the endpoint for publishing a status.
+// Reference: https://docs.joinmastodon.org/methods/statuses/#create
+const mastodonStatusesPath = "/api/v1/statuses"
+
+// mastodonMaxChars is the default status character limit of a Mastodon
+// server. Statuses are truncated to this length as the limit cannot be
+// discovered without an extra API call and exceeding it fails the post.
+const mastodonMaxChars = 500
+
+// Mastodon holds the server URL and OAuth access token
+// for posting statuses to a Mastodon account.
+type Mastodon struct {
+	// URL is the fully resolved statuses endpoint of the server.
+	URL        string
+	ProxyURL   string
+	Token      string
+	Visibility string
+	TLSConfig  *tls.Config
+}
+
+// MastodonPayload is the JSON form accepted by the statuses endpoint.
+type MastodonPayload struct {
+	Status     string `json:"status"`
+	Visibility string `json:"visibility,omitempty"`
+}
+
+// NewMastodon validates the Mastodon server URL and returns a Mastodon
+// object. The address may be the server root URL, in which case the
+// statuses API path is appended. An optional `visibility` query parameter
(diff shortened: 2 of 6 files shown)


xchose:
Rebased onto latest `main`, CI has never run on this PR: could a maintainer approve the run?

xchose:
CI is green now — thanks for approving the run.

The branch is rebased on main and still merges cleanly. Locally `make test` passes with a clean tree after codegen, and I verified the provider end-to-end against a real mastodon.social account.

Could a maintainer take a look when there's time?

matheuscscp:
It feels so strange to send a Flux alert to a social network :thinking: :thinking: :thinking:

stefanprodan:
>It feels so strange to send a Flux alert to a social network

Mastodon can be used like Slack, but self-hosted and private.

docs/spec/v1beta3/providers.md
```diff
@@ -1419,6 +1420,63 @@ stringData:
...
+The [Address](#address) is the Mastodon server root URL, e.g.
+`https://mastodon.social` (the `/api/v1/statuses` path is appended
+automatically). An optional `visibility` query parameter (`public`,
+`unlisted` or `private`) overrides the app's default status visibility,
+e.g. `https://mastodon.social?visibility=unlisted`.
+
+The status text contains the involved object, the event message and the event
+metadata as key-value lines. Statuses longer than 500 characters (the default
+Mastodon server limit) are truncated. An `Idempotency-Key` header derived from
+the event is sent to prevent duplicate statuses on retried requests.
```
matheuscscp, line 1439: I wonder if we really need this since we already do this on our side, or if it would make sense to reuse the same key:

https://github.com/fluxcd/notification-controller/blob/75dc4646047608ebe225a91f71c8547a6c1d7fd8/internal/server/event_server.go#L213-L254

internal/notifier/mastodon.go
```diff
@@ -0,0 +1,144 @@
...
+	if token == "" {
+		return nil, errors.New("empty Mastodon access token")
+	}
+
+	// The visibility is carried as a query parameter of the address
+	// because the Provider API has no dedicated field for it.
+	q := u.Query()
+	visibility := q.Get("visibility")
+	q.Del("visibility")
+	u.RawQuery = q.Encode()
```
matheuscscp, line 79: Could someone mistakenly understand that the mastodon server accepts this query parameter? This solution feels a bit hacky... I can see other providers that don't exactly use a URL also benefiting from more inputs... Can we choose a sensible default here and only expose this input later, after putting more thought into how we can do this properly for all providers?

docs/spec/v1beta3/providers.md
matheuscscp: Since we could do retries it feels like the key can be useful. In this case I wonder if reusing the same key we already have would make sense.

internal/notifier/mastodon.go
xchose: Agreed. Removed: no `visibility` in the payload, statuses follow the account's default posting visibility. Documented in the spec.

docs/spec/v1beta3/providers.md
xchose: Done: the provider now sends the rate-limiter key. The computation moved to `notifier.EventKey` and is passed to the notifier via context, since the token metadata is stripped before dispatch. `eventKeyFunc` is a thin wrapper, tests unchanged.

xchose:
Re-ran the e2e on kind against a real mastodon.social account after the review changes: payload is `{"status": ...}` only, `Idempotency-Key` matches the rate-limiter digest, visibility follows the account setting, and a repeated identical event created no duplicate status.
mastodon-error
mastodon-info
mastodon-profile-scrolled
mastodon-profile

xchose:
Rebased on main, review comments addressed. Could a maintainer approve the CI run?

internal/server/event_handlers.go
```diff
@@ -255,6 +255,11 @@ func (s *EventServer) dispatchNotification(ctx context.Context,
 		pctx, cancel := context.WithTimeout(context.Background(), params.timeout)
 		defer cancel()
 		pctx = notifier.WithAlertMetadata(pctx, alert.ObjectMeta)
+		// Forward the event key computed by eventMiddleware, so that
+		// notifiers share the identity used by the rate limiter.
+		if key, ok := notifier.GetEventKey(ctx); ok {
```
matheuscscp: This is the wrong place to pass down the event key in my opinion. In the event server it makes sense to set it via context, as the code computing the key is a middleware, hence we can't change the interfaces / arguments there. But to downstream providers like mastodon, the event key should be wired like all the other inputs we pass to providers, i.e. via `getNotificationParams`.

matheuscscp: Via `getNotificationParams` and `createNotifier`.

xchose: Done in f7358bd. The key is now read from the request context once in `getNotificationParams`, before the event is mutated, and passed to `createNotifier` as `notifier.WithEventKey`. The Mastodon notifier stores it at construction and sends it as `Idempotency-Key`. The context helpers in the notifier package and the copy into `pctx` are gone; the request context carries the key only inside the event server for the rate limiter.

xchose:
Review comment addressed in f7358bd, the event key is wired through `createNotifier` options like the other notifier inputs.

Re-ran the e2e on kind against a real mastodon.social account with this change: `Idempotency-Key` matches the rate-limiter digest and is stable across repeated identical events, and five dispatches (info, the same error three times, recovery) created exactly three statuses. Could a maintainer approve the CI run?

internal/server/event_handlers.go
```diff
@@ -361,7 +366,7 @@ func (s *EventServer) getNotificationParams(ctx context.Context, event *eventv1.
 		return nil, droppedProviders{}, fmt.Errorf("failed to create commit status: %w", err)
 	}
 
-	sender, token, err := createNotifier(ctx, s.kubeClient, &provider, commitStatus, s.tokenCache)
+	sender, token, err := createNotifier(ctx, s.kubeClient, &provider, commitStatus, eventKey, s.tokenCache)
```
matheuscscp: `createNotifier` already receives the `ctx` so  it can retrieve the event key itself, no need to retrieve it in `getNotificationParams` (sorry, my last comment was a bit misleading)

internal/server/event_server.go
```diff
@@ -210,45 +216,20 @@ func logRateLimitMiddleware(h http.Handler) http.Handler {
...
-
-	tokenKey := fmt.Sprintf("%s/%s", objectGroup, eventv1.MetaTokenKey)
-	token, ok := event.Metadata[tokenKey]
-	if ok {
-		comps = append(comps, "token="+token)
+// eventKeyFromContext returns the event key computed once by eventMiddleware.
+// The key is derived from the event when absent from the context, so that
+// handlers invoked outside the middleware chain still share the same
+// identity. See notifier.EventKey.
+func eventKeyFromContext(ctx context.Context, event *eventv1.Event) string {
```
matheuscscp: We should not fallback to recomputing the event key. As you noted well, the event metadata is mutated while the request is being processed. The `ok` boolean will never be false.

Also, this function (`eventKeyFromContext`) is not necessary, we can use `eventKeyContextKey{}` directly inside `createNotifier`.

matheuscscp (commented):
One last round and we should be good 🙏

internal/server/event_handlers.go
xchose: Done in 2a6f8c5, `createNotifier` reads the key from the context itself and `getNotificationParams` no longer touches it.

internal/server/event_handlers.go
xchose: Done in 2a6f8c5, `createNotifier` now reads `eventKeyContextKey{}` from the `ctx` it already receives and `getNotificationParams` no longer touches the key.

internal/server/event_server.go
xchose: Removed in 2a6f8c5: no fallback and no `eventKeyFromContext`, `eventKeyFunc` returns the key stored by the middleware and `createNotifier` reads the context key directly. `TestEventKeyFunc` now sets the key in the context the same way the middleware does.

matheuscscp (approved):
LGTM

xchose:
Thanks @matheuscscp for the thorough review, I learned a lot about how the event server and the notifiers fit together.

